PR #5, opened 2025-03-06
Topo editor bugs

manishvenu:
Details:
Alper and I discussed that the TopoEditor has a bunch of minor bugs. Here's a list of issues:

- [x] When double clicking, the TopoEditor would (badly) switch views
- [x] When erasing selected basin, the depth view would sometimes show depths below minimum depth (we don't want that)
- [x] Units on the depth axis were weirdly formatted
- [x] Basin selector to have integer labels
- [x] Min depth doesn't work

Is not done:
- Plotting around the topo editor in a different cell plots in the widget, we don't want that! - Couldn't figure out - made issue

Changes:

- Make the only place the graph is changed in refresh_display_mode, and create a trigger_refresh  function to allow actions to redraw the graph as needed. It becomes easier to track how the graph is changing that way
- Add two clim args to the depth mode because that's the way to make sure the vmin works, very happy to take advice on how to make this better
- Force cbar to be integers so it looks better for mask views
- Make min depth max flexible to NaNs

Fixes:
#7

### DIFF
--- a/mom6_bathy/topo_editor.py
+++ b/mom6_bathy/topo_editor.py
@@ -157,8 +157,9 @@ class TopoEditor(widgets.HBox):
         mode = change['new']
 
         if mode == 'depth':
+            self.im.set_clim(vmin = self.topo.min_depth, vmax = self.topo.depth.max(skipna=True).item())
             self.im.set_array(self.topo.depth.data)
-            self.im.set_clim((self.topo.min_depth, self.topo.depth.data.max()))
+            self.im.set_clim(vmin = self.topo.min_depth, vmax = self.topo.depth.max(skipna=True).item()) # For some reason, this needs to be set twice to get the correct minimum bound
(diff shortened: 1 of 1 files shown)


manishvenu:
@AidanJanney, I can't add you as a reviewer, but would appreciate your informal review on this!

AidanJanney:
@manishvenu definitely! I'll pull your branch and take a look as soon as I can!

AidanJanney:
All of the changes are amazing! Switching between the different views, adjusting depth, and everything else works great, but I did find a weird issue with minimum depth that I recorded here: https://github.com/NCAR/mom6_bathy/issues/9#issue-2950928692.

alperaltuntas (approved):
This is great! Other than the ongoing discussion in issue #9, I don't have any additional comments. Feel free to merge this as is, or hold off until the discussion is resolved.

manishvenu:
#9 spun off a few issues that are a bit too much for this scope!

PR #12, opened 2025-05-05
Add Grid Point Masks

manishvenu:
Details:

 For regional cases, I'm interested in calculating the boundary masks, which happen to be on q/u/v points. This is useful so we can figure out if the regridding from source is leaking zeroes into the boundaries. MOM6 calculates the masks, and I put them as properties in the topo object here. 

Changes:

- Mask properties for u, v, and q points. (`qmask`, `umask`, and `vmask`)
- Mask properties combined into `supergridmask`
- Function if lat-lon point on supergrid is ocean or land:`point_is_ocean`

Question:

The q-points are all land on the very corner point (regardless if ocean or not) . I'm not quite sure why, but this is probably a regional thing. Maybe we don't want this in a generic Topo property. Or we call it `regional_qmask`?

Testing:
/glade/u/home/manishrv/documents/croc/dev/bgc/cesm_downscaling/regridding_workflow/test_topo_mask.ipynb

### DIFF
--- a/mom6_bathy/topo.py
+++ b/mom6_bathy/topo.py
@@ -126,21 +126,21 @@ class Topo:
         tmask = self.tmask
 
         # Create empty mask DataArray for umask
-        umask_da = xr.DataArray(
+        umask = xr.DataArray(
             np.ones(self._grid.ulat.shape, dtype=int),
             dims = ['yh','xq'],
             attrs={"name": "U mask"})
         
         # Fill umask with mask values
-        umask_da[:,:-1] &= tmask.values # h-point translates to the left u-point
-        umask_da[:,1:] &= tmask.values # h-point translates to the right u-point
+        umask[:,:-1] &= tmask.values # h-point translates to the left u-point
+        umask[:,1:] &= tmask.values # h-point translates to the right u-point
 
-        return umask_da
+        return umask
     
     @property
     def vmask(self):
         """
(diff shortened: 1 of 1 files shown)


mom6_bathy/topo.py
```diff
@@ -126,6 +193,25 @@ def basintmask(self):
...
+    @property
+    def supergridmask(self):
+        mask_da = xr.DataArray(np.zeros(self._grid._supergrid.x.shape, dtype=int),dims=["nyp", "nxp"])
+        mask_da[::2, ::2] = self.qmask.values
+        mask_da[::2, 1::2] = self.vmask.values
+        mask_da[ 1::2,::2] = self.umask.values
+        mask_da[ 1::2,1::2] = self.tmask.values
+        return mask_da
+
+    def point_is_ocean(self, lon,lat):
```
alperaltuntas, line 206: Could you explain how this method works? Looks like it is for a single point, but returns a list. Also, I am unclear why you are iterating over longitudes.

manishvenu: Added a bit more explanation, the function takes in a list of lons & lats and returns if those points are ocean or land. It iterates over the range of lon to get all the points and then checks them if they are land

manishvenu: Let me know if we should have this function -> if not we can remove. This is definitely something a user can do as well.

alperaltuntas (approved):

PR #15, opened 2025-05-06
Add Pop angle at B-grid u points

manishvenu:
Add POP angle at U-points on B grids to `Grid` and add to `write_cice_grid` in `Topo`.

### DIFF
--- a/mom6_bathy/grid.py
+++ b/mom6_bathy/grid.py
@@ -507,11 +507,11 @@ class Grid:
             attrs={"name": "angle grid makes with latitude line", "units": "degrees"},
         )
 
-        # POP angle
+        # q angle
         self.angle_q = xr.DataArray(
-            sg.angle_dx[2::2, 2::2],
+            sg.angle_dx[::2, ::2],
             dims=["ny", "nx"],
-            attrs={"name": "angle POP u-grid makes with latitude line", "units": "degrees"},
+            attrs={"name": "angle q-grid makes with latitude line", "units": "degrees"},
         )
         # T area
         self.tarea = xr.DataArray(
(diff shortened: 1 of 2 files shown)


mom6_bathy/grid.py
```diff
@@ -505,6 +507,12 @@ def _compute_MOM6_grid_metrics(self):
             attrs={"name": "angle grid makes with latitude line", "units": "degrees"},
         )
 
+        # POP angle
+        self.popangle = xr.DataArray(
```
alperaltuntas, line 511: I suggest we call this `angle_q`. The client (e.g., `write_cice_grid`) may then treat/interpret it as the POP2 angle depending on context.

manishvenu: Ok, that makes sense -- should I include all q angles and have `write_cice_grid` slice it or only the top right q angles like a pop grid would want?

alperaltuntas: We should only include the top right q angles (to be consistent with the rest of mom6_bathy and non-symmetric MOM6). Also, this question makes me wonder whether the indexing is correct. That is: 

(a) `[1::2, 1::2]` for T points, and `[2::2, 2::2]` for Q points.

The actual correct indexing might instead be:

(b) `[::2, ::2]` for T and `[1::2, 1::2]` for Q. 

Could you double check it? One quick check is to look at the shape of `angle_dx`. if its shape is (2 * ny + 1, 2 * nx + 1), the indexing (a) might be correct. Otherwise,  the indexing (b) is probably the correct one.

manishvenu: In write_supergrid, it looks like angle_dx is odd which implies that it is symmetric I think. I thought `Grid` was in general symmetric because qlat/qlat is of different dimensions than ulat/ulat or vlat/vlon

manishvenu: I think we treat the grid as default symmetric because the way we calculate tlon and tlat on `1::2` and qlon as `::2,::2`

alperaltuntas (approved):

PR #16, opened 2025-05-22
Integrate Chlorophyll in mom6_bathy

manishvenu:
1. Add a chlorophyll interpolate and subset based on the grid & topo from a processed seawifs chlorophyll file (provided through CESM inputdata)
2. Add a `fill_interp` file for the `pop_tools` functions used in this workflow
3. Update Util Check

I reproduced the results from the repo on a random panama domain but not from the more recent Gustavo NWA chlorophyll file, I assume they are the same. If need be, I'm happy to check against this to.


mom6_bathy/ocean_fill_and_grid_interp.py
```diff
@@ -0,0 +1,560 @@
...
+                        var[j, i] = var[j, i] + rc * 0.25 * res[j, i]
+                    else:  # do a 1D smooth on pole row
+                        res[j, i] = var[j, ip1] + var[j, im1] - 2.0 * var[j, i]
+                        var[j, i] = var[j, i] + rc * 0.5 * res[j, i]
+
+        res_max = np.max(np.fabs(res)) / np.max(np.fabs(var))
+        iter_cnt += 1
+
+
+def super_sample_grid(ocn_qlat, ocn_qlon, ocn_mask, src_nj, src_ni):
```
alperaltuntas, line 331: I feel like we should provide this feature as a class method within the `Grid` class, e.g., something like:

```
grid = Grid(...)
refined_grid = grid.refine(factor=2)
```

which brings the question of how well this module is integrated with the rest of the `mom6_bathy`. Why don't we use `mom6_bathy` types like `Grid` and `Topo` within this module as opposed to separate numpy arrays like `ocn_qlat, ocn_qlon, ocn_mask` etc.

mom6_bathy/ocean_fill_and_grid_interp.py
```diff
@@ -0,0 +1,560 @@
...
+    w_e = xdist(src_lon[i0], spr_lon) / dx
+    w_w = 1.0 - w_e
+    w_n = ydist(src_lat[j0], spr_lat) / dy
+    w_s = 1.0 - w_n
+    return (w_s * w_w * data[j0, i0] + w_n * w_e * data[j1, i1]) + (
+        w_n * w_w * data[j1, i0] + w_s * w_e * data[j0, i1]
+    )
+
+
+def fill_missing_data(
```
alperaltuntas, line 418: Based on the name of the function, I feel like this should be at most 5-10 lines of function with vectorized numpy or xarray operations as opposed to nested for loops and if statements.

alperaltuntas (changes requested):

mom6_bathy/chl.py
```diff
@@ -0,0 +1,215 @@
...
+from datetime import datetime
+from pathlib import Path
+from .ocean_fill_and_grid_interp import (
+    super_interp,
+    fill_missing_data,
+    super_sample_grid,
+)
+
+
+def gen_chl_empty_dataset(output_path, lon, lat, fill_value=-1.0e34):
```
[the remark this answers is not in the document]
manishvenu: Will do. Originally kept in b/c original method did this as well.

mom6_bathy/ocean_fill_and_grid_interp.py
```diff
@@ -0,0 +1,560 @@
...
+        _iterative_fill_sor(nlat, nlon, var, fillmask, tol, rc, max_iter, ltripole)
+        var[keepNaNs] = np.nan  # Replace NaNs in areas not designated for filling
+    else:
+        fillmask = np.isnan(var) & isvalid_mask
+        _iterative_fill_POP_core(nlat, nlon, var, fillmask, tol, ltripole, max_iter)
+
+    return var
+
+
+@jit(nopython=True, parallel=True)
```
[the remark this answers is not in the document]
manishvenu: Numba is coming as a dependency from some package in the env!

mom6_bathy/ocean_fill_and_grid_interp.py
manishvenu: Done, I haven't integrated any of the global funcs into mom6_bathy. They aren't called expect in the global processing done previously. I can move those functions to those scripts and only keep the ones we are using.

mom6_bathy/chl.py
```diff
@@ -0,0 +1,215 @@
...
+    - The CHL_A variable is filled with the placeholder value -1e34.
+    - The TIME dimension is fixed and marked as unlimited in the NetCDF file.
+    - TIME values represent the approximate midpoint of each month in a climatological year.
+    - This dataset structure mimics that of SeaWiFS chlorophyll climatology products and is intended
+      as a template or placeholder.
+    """
+
+    # === Coordinates ===
+
+    time = np.array(
```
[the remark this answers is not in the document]
manishvenu: I'm not quite sure how this would work, because the CHL file is a one year climatology situation. Would we have one for leap years and one for non leap years? In the meantime, I've added to the docstring this is built with a no leap assumption

manishvenu:
Will this work globally?
Yeah probably, no regional or global assumptions I think.

In fill_interp, the function latlon2ji expects a periodic grid (because of modulo), instead of dealing with that, rewrote it to latlon2ji_simple which enforces that the provided lons must be within the range of the src_lon so we don't have to worry about it. Given that, there shouldn't be any global or regional assumptions anymore.

mom6_bathy/fill_interp.py
```diff
@@ -0,0 +1,153 @@
+import numpy as np
+import xarray as xr
+import scipy
+from mom6_bathy.aux import check_lon_range
+
+
+def latlon2ji(src_lat, src_lon, lat, lon):
```
alperaltuntas, line 7: It looks like this function is very similar to the `get_indices` method in the `Grid` class. The Grid implementation uses a `kdtree`, which is likely much faster and more general-purpose (works with both rectangular and curvilinear grids`. To reduce duplication, could we use `get_indices` instead of `latlon2ji` and `latlon2ji_simple`?

This may not be straightforward, since `get_indices` is currently a method of the `Grid` class and depends on having a `Grid` object. . Ideally, we should use `Grid` objects in the Chlorophyl context as well. But if that’s not feasible, another option might be to refactor `get_indices` into a standalone utility function in `aux.py`, and then call it both from `Grid.get_indices` and from other places that need the same functionality. One challenge with this approach is that we should avoid repeatedly constructing the `kdtree` in some way, as it's computationally expensive.

mom6_bathy/fill_interp.py
```diff
@@ -0,0 +1,153 @@
...
+    nj, ni = len(src_lat), len(src_lon)
+    src_x0 = int((src_lon[0] + src_lon[-1]) / 2 + 0.5) - 180.0
+    j = np.maximum(0, np.floor(((lat + 90.0) / 180.0) * nj - 0.5).astype(int))
+    i = np.mod(np.floor(((lon - src_x0) / 360.0) * ni - 0.5), ni).astype(int)
+    jp1 = np.minimum(nj - 1, j + 1)
+    ip1 = np.mod(i + 1, ni)
+    return j, i, jp1, ip1
+
+
+def latlon2ji_simple(src_lat, src_lon, lat, lon):
```
alperaltuntas, line 17: See my comment for `latlon2ji`.

mom6_bathy/fill_interp.py
```diff
@@ -0,0 +1,153 @@
...
+    jp1 = j + 1
+    ip1 = i + 1
+
+    return j, i, jp1, ip1
+
+
+def super_interp(src_lat, src_lon, data, spr_lat, spr_lon):
+    """
+    Bilinear Interpolation of data from src_lat, src_lon to spr_lat, spr_lon. Does not assume periodicity.
+    """
```
alperaltuntas, line 51: If this function *"does not assume periodicity"*, does that mean it explicitly assumes non-periodicity? In other words, is it still correct for periodic grids, or would it give incorrect results?

mom6_bathy/fill_interp.py
```diff
@@ -0,0 +1,153 @@
...
+    w_s = 1.0 - w_n
+
+    return (w_s * w_w * data[j0, i0] + w_n * w_e * data[j1, i1]) + (
+        w_n * w_w * data[j1, i0] + w_s * w_e * data[j0, i1]
+    )
+
+
+def fill_missing_data(idata, mask, maxiter=0, stabilizer=1.0e-14, tripole=False):
+    """
+    Returns data with masked values objectively interpolated except where mask==0. Does not assume periodicity
```
alperaltuntas, line 78: Same concern: If this function "does not assume periodicity", does that mean it explicitly assumes non-periodicity? In other words, is it still correct for periodic grids, or would it give incorrect results?

mom6_bathy/grid.py
```diff
@@ -650,6 +659,51 @@ def get_indices(self, tlat: float, tlon: float) -> tuple[int, int]:
         j, i = np.unravel_index(indices, self.tlat.shape)
         return int(j), int(i)
 
+    def refine(self, factor: int) -> tuple[np.ndarray, np.ndarray]:
```
alperaltuntas, line 662: Does this `refine` method work correctly with periodic grids where the first and last columns/rows overlap?

mom6_bathy/fill_interp.py
manishvenu: Same thing here!

mom6_bathy/grid.py
manishvenu: Since we just drop the first & last col & row it doesn't work with periodic grids because we don't interpolate between the last row/col & first row/col

manishvenu: Should this be mentioned as something that doesn't work for periodic grids, or just suboptimally?

mom6_bathy/fill_interp.py
manishvenu: This may be a little challenging given lat/lon is a 4D array. But I'll try this and open a new PR!

mom6_bathy/fill_interp.py
alperaltuntas: In that case, I think we should disallow the usage of these functions for periodic grids. The Grid class has a `is_cyclic_x` method that we can utilize to determine whether these functions are called for a periodic case, in which case an error should be thrown. (We probably need a `is_cyclic_y` method as well).

mom6_bathy/fill_interp.py
manishvenu: Gonna need to talk about this!

PR #17, opened 2025-06-02
Add Reproducible Features to Topo Editor

akoweng:
Introduces several improvements to topo_editor.py, making the tool more robust and user-friendly for interactive topography editing. 

**Major Changes:**

1. Undo/Redo Functionality: This includes undo and redo buttons to the TopoEditor class, allowing users to revert or reapply changes (across sessions). Also added a save and load feature to save/load any state made to the topo object from their edits. Added buttons to the control panel.

2. Save/Load Editing States: Added snapshot capabilities. Users can save the current editing state (including edit history) and load it back later, enabling checkpointing. Added buttons to the control panel, with a text box to name a save state.

3. Reset Topo: Added a reset button that restores the topography and minimum depth to its original state. Uses a domain-specific topography saved at initialization. Added a reset button to the control panel.

4. Automatic Logging/Histories: Edit histories are automatically saved and loaded for each domain, stored in 'topo_histories/edit_history_<grid>_<shape>.json'.

I have used this so far on the Panama and 'minimal_demo_three_boundary' low-res domains.

**Future Changes:**

- Implement save states more like a version history, or perhaps like in Photoshop where you can access any point in the edit history simply by clicking.


mom6_bathy/edit_command.py
```diff
@@ -0,0 +1,111 @@
+from abc import ABC, abstractmethod
+import numpy as np
+
+class EditCommand(ABC):
+    @abstractmethod
+    def execute(self, topo):
```
alperaltuntas, line 6: Also, I wonder whether we should replace the `execute` method with the `__call__` method. That way, the `EditCommand` and derived classes would become `callable` objects, such that we could call:

```command()```

instead of ```command.execute()```.

This would also prevent confusing `Command.execute()` vs `CommandManager.execute()` methods. (See `EditHistory` comments.

File: mom6_bathy/command_manager.py [hunk 2]
```diff
@@ -0,0 +1,226 @@
...
+
+class CommandManager(ABC):
+    def __init__(self, domain_id, snapshot_dir="snapshots"):
+        self._undo_history = []
+        self._redo_history = []
+        self.snapshot_dir = snapshot_dir
+        self.domain_id = domain_id
+
+    @abstractmethod
+    def get_domain_id(self):
```
alperaltuntas, line 14: Looks like the PR is shaping up quite nicely, and getting closer to being complete. That said, I’m a bit confused about the overall design and usage of `domain_id`.

From the docstring, it sounds like `domain_id` is intended to be a unique identifier: something like a string, number, or hash. But further in the code, it appears to be a dictionary with multiple key-value pairs. If that's intentional, the naming and documentation could be clarified to avoid confusion. 

There’s also some ambiguity about whether `domain_id` is callable or not. Could you clarify the scenarios where making domain_id callable is desirable, and when it should not be callable?

More broadly, it’s unclear whether `domain_id` is meant to be a general concept that applies beyond the `Topo` class. If so, should it live in the abstract `CommandManager` base class? Or is it tightly coupled with Topo and thus should remain within `TopoCommandManager`?

To move toward a more robust implementation, I suggest first documenting the rationale and intended design:

What problem is domain_id solving?

What should its type and behavior be (e.g., identifier vs. callable vs. data structure)?

How should it relate to other components in the class hierarchy (maybe a quick UML sketch would help)?

At what level should it be implemented: base class, subclass, or something else?

mom6_bathy/command_manager.py
```diff
@@ -0,0 +1,226 @@
...
+        fname = os.path.join(self.snapshot_dir, f"{name}.json")
+        data = {
+            "domain_id": self.get_domain_id(),
+            "undo_history": [cmd.serialize() for cmd in self._undo_history],
+            "redo_history": [cmd.serialize() for cmd in self._redo_history]
+        }
+        with open(fname, "w") as f:
+            json.dump(data, f)
+
+    def load_commit(self, name, command_registry, topo):
```
alperaltuntas: Same goes for this method.

mom6_bathy/command_manager.py
```diff
@@ -0,0 +1,237 @@
...
+        with open(fname, "w") as f:
+            json.dump(data, f)
+
+    def load_commit(self, name, command_registry, topo, reset_to_golden=False):
+        """
+        Loads a named snapshot/commit.
+        If reset_to_golden is True, resets topo to golden/original state before applying undo history.
+        If False (default), applies undo/redo history on top of current in-memory topo.
+        """
+        if reset_to_golden:
```
alperaltuntas, line 160: What is golden (or golden state)?

mom6_bathy/topo_editor.py
```diff
@@ -2,30 +2,179 @@
...
+            self.construct_interactive_plot()
+            self.construct_observances()
+            self.initialize_history()
+            self.refresh_commit_dropdown()
+            super().__init__([self._control_panel, self._interactive_plot])
+        else:
+            super().__init__([])
+
+    def initialize_history(self):
+        self.command_manager.execute(InitializeHistoryCommand(self.command_manager, COMMAND_REGISTRY, self.topo))
```
alperaltuntas, line 59: With the elimination of `InitializeHistoryCommand`, this call would be simplified as:

```
self.command_manager.initialize_history()
```

Notice how generic it becomes. This applies to the calls involving other classes that I think should be eliminated (e.g., `UndoCommand`, `RedoCommand`, etc.

alperaltuntas (changes requested):

mom6_bathy/topo_editor.py
```diff
@@ -2,30 +2,187 @@
...
+        self.command_manager.initialize()
+        self.update_undo_redo_buttons()
+
+    def get_topo_id(self):
+        grid = self.topo._grid
+        grid_name = getattr(grid, "name", getattr(grid, "_name", None))
+        shape = [int(v) for v in self.topo.depth.data.shape]
+        return {"grid_name": grid_name, "shape": shape}
+
+    def _ensure_golden_topo(self):
```
manishvenu, line 65: We have not done a good job of this throughout the package, but I'd say most of these functions should have docstrings. Ensure Golden Topo is a pretty cryptic name, so a quick explanation would be great.

mom6_bathy/topo_editor.py
```diff
@@ -2,30 +2,187 @@
...
+        grid_name = topo_id['grid_name']
+        shape = topo_id['shape']
+        shape_str = f"{shape[0]}x{shape[1]}"
+        
+        golden_dir = self.GOLDEN_DIR
+        os.makedirs(golden_dir, exist_ok=True)
+        golden_topo_path = os.path.join(golden_dir, f"golden_topo_{grid_name}_{shape_str}.npy")
+        golden_min_depth_path = os.path.join(golden_dir, f"golden_min_depth_{grid_name}_{shape_str}.json")
+        if not os.path.exists(golden_topo_path):
+            np.save(golden_topo_path, self.topo.depth.data)
```
manishvenu, line 76: Is there a reason to save as npy vs NetCDF? Efficiency or smthng?

mom6_bathy/topo_editor.py
```diff
@@ -2,30 +2,187 @@
...
+        self.command_manager.undo()
+        self.update_undo_redo_buttons()
+        self.trigger_refresh()
+    
+    def redo_last_edit(self, b=None):
+        self.command_manager.redo()
+        self.update_undo_redo_buttons()
+        self.trigger_refresh()
+
+    def on_reset(self, b=None):
```
manishvenu, line 128: Would change to verb first like undo & redo!
```suggestion
    def reset(self, b=None):
```

mom6_bathy/topo_editor.py
```diff
@@ -2,30 +2,187 @@
...
+        self.update_undo_redo_buttons()
+        print("Topo reset to original state.")
+
+    def update_undo_redo_buttons(self):
+        if hasattr(self, "_undo_button"):
+            self._undo_button.disabled = not bool(self.command_manager._undo_history)
+        if hasattr(self, "_redo_button"):
+            self._redo_button.disabled = not bool(self.command_manager._redo_history)
+
+    def refresh_commit_dropdown(self):
```
manishvenu, line 146: We'll have to talk about this in our git discussion!

mom6_bathy/topo_editor.py
```diff
@@ -139,31 +295,141 @@ def construct_control_panel(self):
...
+            layout={'width': '90%'}
+        )
+        self._git_checkout_button = widgets.Button(description='Checkout', layout={'width': '44%'})
+       
+        self._git_merge_source_dropdown = widgets.Dropdown(
+            options=git_list_branches(CROCODASH_REPO_ROOT),
+            description='Merge from:',
+            layout={'width': '90%'}
+        )
+        self._git_merge_button = widgets.Button(description='Merge Branch', layout={'width': '44%'})
```
manishvenu, line 348: You have any questions on the layout for the UI?

mom6_bathy/topo_editor.py
```diff
@@ -175,107 +441,266 @@ def refresh_display_mode(self, change):
...
-                self.topo.depth.data[j, i] = change['new']
-                self.trigger_refresh()
-
-        self._depth_specifier.observe(
-            on_depth_change,
-            names='value',
-            type='change'
-        )
-
+    # --- Git Callbacks ---
```
manishvenu, line 598: I wonder if all of this more generic git stuff can be moved to a seperate, generic file.

mom6_bathy/topo_editor.py
```diff
@@ -175,107 +441,266 @@ def refresh_display_mode(self, change):
...
+
+    def on_git_checkout(self, b):
+        target = self._git_branch_dropdown.value
+        if not target:
+            print("Please select a branch to checkout.")
+            return
+        try:
+            rel_snapshot_dir = os.path.relpath(os.path.abspath(self.SNAPSHOT_DIR), CROCODASH_REPO_ROOT)
+            success, untracked, changed, staged = git_safe_checkout_branch(CROCODASH_REPO_ROOT, target, rel_snapshot_dir)
+            if not success:
```
manishvenu, line 658: Maybe this whole print statement after git_safe_checkout_branch should go into the function itself.

mom6_bathy/topo_editor.py
```diff
@@ -175,107 +441,266 @@ def refresh_display_mode(self, change):
...
+
+            self._git_branch_dropdown.options = git_list_branches(CROCODASH_REPO_ROOT)
+            self._git_branch_dropdown.value = git_get_current_branch(CROCODASH_REPO_ROOT)
+            self._git_merge_source_dropdown.options = git_list_branches(CROCODASH_REPO_ROOT)
+
+            self.on_reset()
+
+            snapshots = [f for f in os.listdir(self.SNAPSHOT_DIR) if f.endswith('.json') and not f.startswith('golden_')]
+            if snapshots:
+                snapshots.sort(key=lambda f: os.path.getmtime(os.path.join(self.SNAPSHOT_DIR, f)), reverse=True)
```
manishvenu, line 677: This sorts by time?Maybe quick comments here would help

mom6_bathy/git_utils.py
```diff
@@ -0,0 +1,143 @@
...
+            os.makedirs(os.path.dirname(abs_path), exist_ok=True)
+            with open(abs_path, "w") as f:
+                f.write(file_contents)
+            return True, f"Restored '{os.path.basename(file_path)}' to commit {commit_sha}."
+        except git.exc.GitCommandError as e:
+            return False, str(e)
+    else:
+        raise ValueError(f"Unknown action: {action}")
+
+def git_commit_info(repo_root, rel_dir=None, commit_sha=None, file_path=None, mode='list', branch=None):
```
manishvenu, line 76: Why do all these function names from here & below start with git? Probably don't need that

manishvenu (commented):
A few comments - very neat body of work - seems to boil down to two ideas, git & command pattern.

PR #22, opened 2025-06-24
Add an example usage of the new mapping module

alperaltuntas:
This PR adds a notebook demonstrating how to use the mapping module of mom6_bathy to create an ESMF mapping file from a runoff source grid (JRA025) to a target MOM6 grid (tx2_3v2).


alperaltuntas:
TODO: steps 2 and 3 should be combined and reduced to a single mom6_bathy API call, as opposed to a series of sparse matrix operations and multiple API calls.

manishvenu:
This to, so sorry, will recreate!

PR #23, opened 2025-06-26
Create VGrid objects from cell center and cell interface.

AidanJanney:
Introduces new methods to load a VGrid from files with different information about a vertical grid. 

**Major Changes:**
- Added new functionality to `VGrid.from_file()` – originally it took a file path as an argument, and it assumed this netCDF file contained a variable "dz" describing layer thickness for a vertical grid. This new implementation retains this default behavior, but it allows users to pass in a different variable name and specify two additional types of information for creating the vertical grid: cell center depth and cell interface depth.

- Added attribute to get cell interface depth from a VGrid object. 

- Added tests to for VGrid class. 

**Minor Changes:**
- Added some checks to ensure `dz` has no zero values
- Update the example data for tests. It would be more useful for it to test inputs like negative depths.

**Questions:**
- Cell center and cell interface calculations assume the grid always has an interface at depth = 0. Is this reasonable?
- The tests allow for some floating point error using `np.allclose` when translating cell center/interface to layer thickness and vice versa – is this an acceptable allowance?


AidanJanney:
@manishvenu and @alperaltuntas I would love any feedback on this! The commit history might be a little funky because I messed up a merge, but only one existing file was changed: `vgrid.py` and only two files created `/tests/conftest.py` and `/tests/test_vgrid.py`.

manishvenu (approved):
Great!

AidanJanney:
@manishvenu thank you! put in both of your suggestions.

manishvenu:
I accdeintally closed this Aidan, deleting the CrocoDash branch so sorry!

PR #24, opened 2025-06-27
Update Python Env

manishvenu:
Details:
Update Python in mom6_bathy so we no longer have a CrocoDash branch.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ channels:
   - anaconda
   
 dependencies:
-  - python>=3.11.10,<3.12
+  - python>=3.12.7,<3.13.0
   - xesmf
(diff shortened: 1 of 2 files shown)


alperaltuntas:
> Why do we have 28 commits here?

Actually, never mind. I'll squash and merge. So these phantom commits should not matter (I hope).

manishvenu:
> > Why do we have 28 commits here?
> 
> 
> 
> Actually, never mind. I'll squash and merge. So these phantom commits should not matter (I hope).

Yeah! It was some mistake with the merge, but, same, figured it wouldn't matter.

alperaltuntas (approved):

PR #25, opened 2025-06-30
Create VGrid objects from cell center and cell interface.

AidanJanney:
Reviving PR https://github.com/NCAR/mom6_bathy/pull/23 after some changes to mom6_bathy structure.

Introduces new methods to load a VGrid from files with different information about a vertical grid. 

**Major Changes:**
- Added new functionality to `VGrid.from_file()` – originally it took a file path as an argument, and it assumed this netCDF file contained a variable "dz" describing layer thickness for a vertical grid. This new implementation retains this default behavior, but it allows users to pass in a different variable name and specify two additional types of information for creating the vertical grid: cell center depth and cell interface depth.

- Added attribute to get cell interface depth from a VGrid object. 

- Added tests to for VGrid class. 

**Minor Changes:**
- Added some checks to ensure `dz` has no zero values
- Update the example data for tests. It would be more useful for it to test inputs like negative depths.

**Questions:**
- Cell center and cell interface calculations assume the grid always has an interface at depth = 0. Is this reasonable?
- The tests allow for some floating point error using `np.allclose` when translating cell center/interface to layer thickness and vice versa – is this an acceptable allowance?


AidanJanney:
Very weird that it shows so many commits, but only three files are edited.

manishvenu (approved):

PR #26, opened 2025-07-01
Integrate CHL into mom6_bathy

manishvenu:
See previous PR: https://github.com/NCAR/mom6_bathy/pull/16


manishvenu:
@alperaltuntas Definitely would like to talk through this in a meeting! It's diverged heavily from the source code.

PR #42, opened 2025-10-30
Try Packaging - TestPyPi on pushes to main, PyPi on release - Conda forge is a one & done (done manually)

manishvenu:
1. Update README
2. Bug Fixes for Tests
3. Github Action to push to testpypi on new versions and pypi on new releases. (Conda Forge is being added as well, https://github.com/conda-forge/staged-recipes/pull/32909)

It is already available on testpypi/pypi


alperaltuntas (commented):
This is a great idea. However, I am not sure whether we should do this now. Months ago we floated the idea of renaming mom6_bathy to something more descriptive. I wonder whether we should wait for the renaming to happen. If you think this would be useful in the short term, though, feel free to merge it.

alperaltuntas (approved):

PR #56, opened 2026-03-13
(Regional) Projected Supergrids (Curvilinear Generation Support)

### LINKED ISSUE #195: Projected Grid Creation

manishvenu:
Closes CROCODILE-CESM/CrocoDash#195

Details:
Lots of things in this PR, and was heavily supported with Claude, but mainly adding support to create curvilinear grids for primarily two use cases: Artic/Antartic Grids & estuaries (as in rotating the grid so that only one boundary is at the mouth). This pairs with the interactive grid PR because doing this without a map is hard. 

Changes:
1. Introduces a new supergrid class called ProjectedSupergrid that has two methods:  

- from_crs -> name a projection and extents in distance units
- from_center -> name a center point, rotation, and distance extents

2. Add grid params attribute to supergrid datasets so they can be recreated very directly
3. Add a from_ds method on supergrid base class so that Grid class can use that instead of the assignment of variables by creating a dummy grid first

Minor Changes:
1. Fix a small issue with TopoLibrary where it shows up in git


manishvenu:
@alperaltuntas This is ready for review now after the rotation code was merged.

manishvenu:
#58 is companion

mom6_forge/topo.py
```diff
@@ -55,6 +48,7 @@ def __init__(self, grid, min_depth, version_control_dir="TopoLibrary"):
 
         # Create a folder to store bathymetry objects in
         self.topos_root = Path(version_control_dir).mkdir(exist_ok=True)
+        (Path(version_control_dir) / ".gitignore").write_text("*")
```
alperaltuntas: Is this change compatible with the latest git/version control changes? Also, when I attempted to merge this into the latest main locally, I encountered conflicts with git/version control changes. Please merge the latest main into this branch and take extra care in resolving those conflicts.

mom6_forge/_supergrid.py
```diff
@@ -53,6 +56,9 @@ def __init__(self, x, y, dx, dy, area, angle_dx, axis_units):
             Local grid angle relative to east.
         axis_units : str
             Units of x and y (e.g. "degrees" or "meters").
+        grid_params : dict
```
alperaltuntas: The argument name `grid_params` sound overly general, if this dictionary contains only dataset attributes. Also, if `grid_type` is required, then I suggest we make it an argument rather than specifying it in `grid_params`

mom6_forge/_supergrid.py
```diff
@@ -241,18 +268,34 @@ def from_extents(
...
+                ny=ny,
+                radius=radius,
+            ),
+        )
 
     @classmethod
-    def from_xy(cls, x, y, radius=_DEFAULT_RADIUS):
+    def from_xy(cls, x, y, radius=_DEFAULT_RADIUS, grid_params=None):
         """Create a grid directly from coordinate arrays."""
+        if grid_params is None:
```
alperaltuntas: Why not just set the default value of `grid_params` to `{}` instead of `None` and get rid of this if block??

mom6_forge/_supergrid.py
```diff
@@ -336,6 +396,236 @@ def _build_grid(self, lon_min, len_x, lat_min, len_y, resolution, radius):
...
+            Geographic coordinates of all supergrid nodes in degrees.
+        grid_params : dict
+            Construction parameters to store on the instance.
+        radius : float, optional
+            Sphere radius in metres. Defaults to Earth's IUGG mean radius.
+        """
+        # Clamp to valid geographic range (floating-point overshoot from projection)
+        y = np.clip(y, -90.0, 90.0)
+
+        # dx, dy, area: use base class consistent calcuyion methods
```
alperaltuntas: typo.

mom6_forge/_supergrid.py
```diff
@@ -336,6 +396,236 @@ def _build_grid(self, lon_min, len_x, lat_min, len_y, resolution, radius):
...
+        dx, dy = SupergridBase._calc_dx_dy(x, y, R=radius)
+        area = SupergridBase._calc_area(x, y, R=radius)
+
+        # angle_dx: angle of grid i-direction reyive to east
```
alperaltuntas: typo.

mom6_forge/_supergrid.py
```diff
@@ -336,6 +396,236 @@ def _build_grid(self, lon_min, len_x, lat_min, len_y, resolution, radius):
...
+
+    Parameters
+    ----------
+    ds : xr.Dataset
+        A supergrid dataset written by SupergridBase.to_ds (or Grid.write_supergrid).
+
+    Returns
+    -------
+    type
+        One of UniformSphericalSupergrid, RectilinearCartesianSupergrid,
```
alperaltuntas: This docstring seems incorrect and contradicts the function signature (`str | None`)/

mom6_forge/_supergrid.py
```diff
@@ -336,6 +396,236 @@ def _build_grid(self, lon_min, len_x, lat_min, len_y, resolution, radius):
...
+        # angle_dx: angle of grid i-direction reyive to east
+        # shape: (2*ny+1, 2*nx+1)
+        angle_dx = SupergridBase.calc_supergrid_rotation_angles_using_expanded_supergrid_method(
+            x, y
+        )
+        grid_params["radius"] = radius
+        return cls(x, y, dx, dy, area, angle_dx, "degrees", grid_params)
+
+
+_GRID_TYPE_TO_CLASS = {
```
alperaltuntas: This doesn't seem to be used anywhere. If unnecessary, please remove.

mom6_forge/_supergrid.py
```diff
@@ -336,6 +396,236 @@ def _build_grid(self, lon_min, len_x, lat_min, len_y, resolution, radius):
...
+        A supergrid dataset written by SupergridBase.to_ds (or Grid.write_supergrid).
+
+    Returns
+    -------
+    type
+        One of UniformSphericalSupergrid, RectilinearCartesianSupergrid,
+        ProjectedSupergrid, or SupergridBase (fallback for datasets without a
+        grid_type attribute).
+    """
+    if type(ds) is str:
```
alperaltuntas: Use `isinstance` for type checking.

alperaltuntas (commented):
I left a few line-specific comments, but here is a more general observation:
 
 Saving construction parameters (lon_min, crs_wkt, center_lat, etc.) into the grid file's global attributes and relying on them for object creation seems fragile to me. It works today, but I don't think it is maintainable or extensible. Things will drift and become incompatible over time.

For instance, if a keyword gets renamed, every old grid file will fail to load with no indication of a version/schema mismatch. And how do we handle adding new parameters? Also, general metadata like Created and Author is mixed with construction parameters in the same flat namespace. On the next save, these values are written again, which will overwrite fresh values with old ones.

I might be wrong, but the data needed to rebuild a `SupergridBase` is already in x, y, dx, dy, area, and angle_dx. `ProjectedSupergrid._from_xy` already shows that the metrics can be recomputed from `x` and `y` alone. So `from_ds` should rebuild from the arrays. Only `grid_type` is needed as an attribute to choose the correct subclass. And as I suggest in the line comments, perhaps it should be a separate argument rather than living in ds.attrs at all. The construction parameters might still be saved, but only as an optional hint.

mom6_forge/topo.py

mom6_forge/_supergrid.py
manishvenu: Done

mom6_forge/_supergrid.py
manishvenu: Done

mom6_forge/_supergrid.py
manishvenu: Done

mom6_forge/_supergrid.py
manishvenu: Done

mom6_forge/_supergrid.py
manishvenu: Done

mom6_forge/_supergrid.py
manishvenu: Done

manishvenu:
Ok, in response to the grid params comment, the reason I was saving the grid parameters was that it is challenging to back out the exact arguments used in the init for things like projected grids. I wanted those so it's easy to get them in the grid_creator (#58).

I'll see if I can find a better solution.

alperaltuntas:
> Ok, in response to the grid params comment, the reason I was saving the grid parameters was that it is challenging to back out the exact arguments used in the init for things like projected grids. I wanted those so it's easy to get them in the grid_creator (#58).
> 
> I'll see if I can find a better solution.

Thanks! Please let me know when it's ready for another review.

mom6_forge/grid.py
alperaltuntas: Have you tested if this still works?

mom6_forge/_supergrid.py
```diff
@@ -13,9 +15,10 @@
 import xarray as xr
 from datetime import datetime
 from typing import Optional
-from mom6_forge.utils import normalize_deg
 
 _DEFAULT_RADIUS = 6.371e6  # mean radius of the Earth (IUGG), in metres
+from pyproj import CRS, Transformer
```
alperaltuntas, line 20: Do we need to add pyproj to `pyproject.toml`?

mom6_forge/_supergrid.py
```diff
@@ -150,8 +214,26 @@ def to_ds(self, name=None, author: Optional[str] = None) -> xr.Dataset:
...
+        return cls(
+            ds.x.data,
+            ds.y.data,
+            ds.dx.data,
+            ds.dy.data,
+            ds.area.data,
+            ds.angle_dx.data,
+            ds.x.attrs.get("units", "degrees"),
+            grid_type=ds.attrs.get("grid_type"),
+            radius=ds.attrs["radius"],
```
alperaltuntas, line 232: Should we have a fallback if `radius` attribute doesn't exist?

mom6_forge/grid.py
```diff
@@ -463,6 +464,117 @@ def get_bounding_boxes_of_rectangular_grid(cls, hgrid):
             "ic": init_result,
         }
 
+    @classmethod
+    def from_projection(
```
alperaltuntas, line 468: please add tests for this method.

mom6_forge/grid.py
```diff
@@ -463,6 +464,117 @@ def get_bounding_boxes_of_rectangular_grid(cls, hgrid):
...
+            ystart=float(np.min(sg.y)),
+            name=name,
+        )
+        obj._supergrid = sg
+        obj._compute_MOM6_grid_metrics()
+        obj.grid_type = "projected_crs"
+        return obj
+
+    @classmethod
+    def from_center(
```
alperaltuntas, line 508: Please add tests for this method. (There are tests for ProjectedSupergrid.from_center, but not for this one).

alperaltuntas (changes requested):

mom6_forge/_supergrid.py
manishvenu: I don' t think so -> I think you need radius to calculate dx/dy/area

mom6_forge/grid.py
manishvenu: Done

mom6_forge/grid.py
manishvenu: Done

mom6_forge/grid.py
alperaltuntas: More specifically, I am wondering if `Grid(type="rectilinear_cartesian", …)` still works. If there isn't a test for this, please add one.

manishvenu: Oh shoot, missed this comment will do.

mom6_forge/_supergrid.py
manishvenu: pyproj comes with cartopy!

alperaltuntas:
@manishvenu there are still a couple of unresolved comments above.

mom6_forge/grid.py
manishvenu: Added a test and fixed the associated bug.

alperaltuntas (approved):

PR #58, opened 2026-03-16
Interactive Grid Editor refactor w/ Point & Click

### LINKED ISSUE #194: Interactive Grid Creation (Point & Click)

manishvenu:
Details:
This is a way to create regional grids with projections in the UI using the new features in proj_supergrid branch.

Closes CROCODILE-CESM/CrocoDash#194

Changes:
1. Add a "create" mode to the grid editor for latlon or projected grids
2. You cannot edit projected grids that come from outside the GridCreator


alperaltuntas (commented):
Two minor comments. Otherwise, LGTM

manishvenu:
I'm using CLaude for this and it's working pretty well.

mom6_forge/grid_creator.py
```diff
@@ -72,164 +248,514 @@ def construct_control_panel(self):
...
-                self._commit_details,
-                widgets.HBox([self._save_button, self._load_button]),
-            ]
-        )
+    def _switch_to_grid_mode(self):
+        """Replace the creator controls panel after a grid is created or loaded."""
+        creator_controls = self._build_creator_controls()
+        library_section = self._control_panel.children[1]
+        self._control_panel.children = [creator_controls, library_section]
+        self.construct_observances()
```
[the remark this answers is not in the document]
alperaltuntas: @manishvenu Coming back to this issue, why do we need to construct observances every time the grid mode is switched. Can you get to the bottom of this? 

> If I remove this, the controls aren't responsive anymore!

To me, this indicates a latent bug then.

mom6_forge/grid_creator.py
```diff
@@ -392,30 +1041,33 @@ def _on_slider_change(self, change):
...
+            self._build_creator_controls(),
+            self._control_panel.children[1],
+        ]
+        self.construct_observances()
+        if not isinstance(self._current_map_proj, ccrs.PlateCarree):
+            self._set_map_projection(ccrs.PlateCarree(), None)
+        else:
+            self.plot_world()
+        self._start_click_mode()
+        return
```
alperaltuntas: What's the purpose of the return keyword here?

mom6_forge/grid_creator.py
```diff
@@ -392,30 +1041,33 @@ def _on_slider_change(self, change):
...
+        self._control_panel.children = [
+            self._build_creator_controls(),
+            self._control_panel.children[1],
+        ]
+        self.construct_observances()
+        if not isinstance(self._current_map_proj, ccrs.PlateCarree):
+            self._set_map_projection(ccrs.PlateCarree(), None)
+        else:
+            self.plot_world()
+        self._start_click_mode()
```
alperaltuntas, line 1119: In this method, do we need to reset `_mode_selector` and `layout.display` too?

mom6_forge/grid_creator.py
```diff
@@ -5,26 +5,89 @@
...
-        self._initial_params = {
-            "lenx": grid.lenx,
-            "leny": grid.leny,
-            "nx": grid.nx,
-            "ny": grid.ny,
-            "xstart": grid.supergrid.x[0, 0],
-            "ystart": grid.supergrid.y[0, 0],
-            "name": grid.name,
-        }
+        (self.grids_dir / ".gitignore").write_text("*\n")
```
alperaltuntas, line 78: This looks unsafe. Could you describe the intent here?

alperaltuntas (changes requested):

mom6_forge/grid_creator.py
manishvenu: So that I don't commit my grids in the Grid Library to my git repo

alperaltuntas: Which git repo?

manishvenu: If I had a notebook I was messing with, I don't want the git commit for my notebook to accidentally add the GridLibrary to my remote.

mom6_forge/grid_creator.py
manishvenu: None, removed

mom6_forge/grid_creator.py
manishvenu: Not necessarily because the mode can remain what the user said, it'll just go back to the global map and be on that mode but in "create" style.

mom6_forge/grid_creator.py
alperaltuntas: I am not following. I am still unsure which git repo you are talking about. Is this *notebook you are messing with* part of the mom6_forge repo, or some external repo?

manishvenu: An External Repo->

As a user, I'd have a notebook in a folder. That would generate GridLibrary, TopoLibrary, etc...

If I'm commiting this notebook to a git repo, the GridLibrary shouldn't have its grids commited as well.

mom6_forge/grid_creator.py
alperaltuntas: What if the user was in "From Projection" mode with the polar stereographic map active, then hits Reset? The map gets forced back to PlateCarree, but _mode_selector.value is still "From Projection" Isn't that a broken state?

manishvenu: The default for projection mode is still plate caree, but I'll check that right now

manishvenu: It's working!

alperaltuntas: Could you double-check the specific sequence: pick EPSG:3995 from the dropdown so the map switches to NorthPolarStereo, then hit Reset without creating a grid first. Does the map stay polar, or get back to PlateCarree while the dropdown still reads EPSG:3995?

manishvenu: You can't click reset without creating a grid first, but yes if you create a grid in northpolar and then reset the dropdown doesn't reset back to platecaree I will add a reset for the dropdown as well.

mom6_forge/grid_creator.py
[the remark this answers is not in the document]
alperaltuntas: @manishvenu this is the only remaining unresolved comment for this pr.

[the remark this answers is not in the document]
manishvenu: Sorry, forgot about this one give me a few minutes!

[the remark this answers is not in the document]
manishvenu: Should be removed, in the original design was the objects (sliders) were being reinitialized every time, so the observances needed to be redone. That's no longer the case anymore.

alperaltuntas (approved):

PR #62, opened 2026-03-24
Rename mom6_bathy as mom6_forge

alperaltuntas:
This PR renames the `mom6_bathy` library as `mom6_forge` since the feature list now includes more than grid and bathymetry generation, and will likely to expand. The default conda env name has also been changed from `mom6_bathy_env` to `mom6_forge`. The repo name will also change accordingly upon the merge of this PR.


manishvenu (approved):
It's running for me! Sure we'll find something later that we'll need to change!

PR #64, opened 2026-04-07
Move rotation from regional_mom6 to mom6_bathy and move metric calculation into supergridbase

manishvenu:
Details:

1. Add rotation from regional_mom6 to  _supergrid.py base class
2. the new module is for metric calculation for the supergrid (utils is too big)
3. If you solve the dx/dy in midas, the basic form we need is the exact same as regional_mom6, so I moved this to the base class
4. Quadrilateral Areas is now used in the "midas" method, so it's no longer exactly reproducible to the midas method.


manishvenu:
@ashjbarnes Just an fyi this is here!

ashjbarnes:
yeah lgtm!

tests/test_rotation.py
```diff
@@ -0,0 +1,360 @@
...
+    # Check Size
+    assert len(expanded_supergrid.nxp) == (len(supergrid.nxp) + 2)
+    assert len(expanded_supergrid.nyp) == (len(supergrid.nyp) + 2)
+
+    # Check pseudo_supergrid keeps the same values
+    assert (expanded_supergrid.x.values[1:-1, 1:-1] == supergrid.x.values).all()
+    assert (expanded_supergrid.y.values[1:-1, 1:-1] == supergrid.y.values).all()
+
+    # Check extra boundary has realistic values
+    assert (
```
alperaltuntas, line 32: Please use abs values in these comparisons to catch negative differences with large magnitudes.

manishvenu:
@alperaltuntas This is ready for your review again! Made the changes we discussed.

mom6_forge/grid.py
```diff
@@ -519,6 +520,47 @@ def from_supergrid_ds(cls, ds: xr.Dataset, name: Optional[str] = None) -> "Grid"
...
+
+        Returns
+        -------
+        xr.DataArray
+            The t-point angles
+        """
+        # t-points: cell centers at every other supergrid point starting at index 1
+        t_points = xr.Dataset(
+            {
+                "x": (("nyp", "nxp"), self.tlon.values),
```
alperaltuntas: why are the dimensions nyp and nxp? Shouldn't they be ny and nx?

mom6_forge/grid.py
```diff
@@ -519,6 +520,47 @@ def from_supergrid_ds(cls, ds: xr.Dataset, name: Optional[str] = None) -> "Grid"
...
+    def calculate_t_point_rotation_angles_from_mom6_method(self) -> xr.DataArray:
+        """
+        Calculate the ``angle_dx`` in degrees from the true ``x`` direction (parallel to latitude) counter-clockwise
+        and return as a dataarray. (Mimics MOM6 angle calculation function :func:`~mom6_angle_calculation_method`)
+
+        Use this to compare with the angle_dx provided in the supergrid file to check if the angles are consistent with MOM6 method
+
+        Parameters
+        ----------
+        supergrid: xr.Dataset
```
alperaltuntas: Seems like this Parameter doc should be removed.

alperaltuntas:
I left a couple of small comments. Once those are sorted, this is good to merge. Also please make sure to tidy up the commit message before squash and merge. Thanks!

manishvenu:
Hey @alperaltuntas, There are no comments visible!

Thanks,
Manish

alperaltuntas (commented):
We already compute `angle_dx` in _supergrid.py.

mom6_forge/grid.py
manishvenu: Yup, removed this because we already calc angle_dx in supergrid

mom6_forge/grid.py
manishvenu: Yup, removed this because we already calc angle_dx in supergrid

PR #65, opened 2026-04-09
Initial commit of git=False option to massively speed up topo processing

briandobbins:
The _purpose_ of this PR is to add a flag to creating topography files of "git=False", which has massive speed-ups for large domains -- some comparative data below when using Derecho's scratch file system, which is a tad terrible with small-file operations like Git updates.  Since I'm a total novice with Python, I used Claude to help write this, so it deserves a lot of healthy skepticism from any review, but it's a conceptually simple change, at least.

I kept the default behavior to of course keep the Git integration set to true since that's necessary for change-tracking in CROCODILE workflows, but now either 'git=False' or 'version_control_dir=None' should avoid creating the (default) TopoLibrary directory.

| Topo.from_topo_file time (s) | git=True (default) | git=False |
|:----------------------------|:-----------------:|:---------:|
| 2/3rds grid   | 5.6 | 0.4 |
| 1/12th grid | 203.3 | 16.2 |
| 1/36th grid | 30803.0 | 140.6 |

I might have another PR later, but this is the big one, and I'd love to hear feedback on this since I don't know how it fits within the CROCODILE / mom6_forge goals.  I also think this shouldn't effect the widget-based usage, which always uses Git, but this is super helpful for fast grid generation on large domains.


briandobbins:
Note, too, that I'm contributing this here based on the workflow that was suggested to me for creating the 1/36th degree grid, which uses this tool.  That example was here:

https://ncar.github.io/tx2_3/mesh/create_mesh.html

manishvenu (commented):
Hey Brian,

Thanks for this! This is super helpful, there's a few things we likely want to add/change:

1. A test with git off
2. A check in the topo_editor.py file that the tcm is enabled
3. the `version_control_dir is None` shouldn't need to be kept as legacy code, it's pretty new

Do you want me to take it from here? I'm happy to, especially if you're busy.

Thanks,
Manish V.

briandobbins:
Sure, if you don't mind taking it from here, it's all yours.   I did do _one_ test, offline, generating with Git on and Git off, and comparing the NetCDF files, and they matched, but any additional checks you feel are warranted are great.

Thank you so much!

manishvenu:
> Sure, if you don't mind taking it from here, it's all yours. I did do _one_ test, offline, generating with Git on and Git off, and comparing the NetCDF files, and they matched, but any additional checks you feel are warranted are great.
> 
> Thank you so much!

Great, thanks a bunch for this! Will do

manishvenu:
Hey @alperaltuntas, 

This PR is ready for review! (Ignore docs failing, tests are passing)

mom6_forge/topo.py
alperaltuntas: Here, you need to add `git=self.version_control`

mom6_forge/topo.py
```diff
@@ -32,6 +31,14 @@ def __init__(self, grid, min_depth, version_control_dir="TopoLibrary"):
             horizontal grid instance for which the bathymetry is to be created.
         min_depth: float
             Minimum water column depth. Columns with shallow depths are to be masked out.
+        version_control_dir: str, optional
+            Directory in which to store version-controlled bathymetry data. Defaults to
+            "TopoLibrary". Ignored if git is False (version control is no longer used)
+        git: bool, optional
+            If True (default), initialize a git repository for version control and
+            undo/redo support. If False, skip all git and filesystem side-effects;
+            note that save(), undo(), redo(), branch, and tag operations will be
```
alperaltuntas: Since `save()` is now removed, isn't it unavailable regardless? Also, why was `save()` removed?

mom6_forge/topo.py
```diff
@@ -46,34 +53,39 @@ def __init__(self, grid, min_depth, version_control_dir="TopoLibrary"):
         self._min_depth = min_depth
         self.land_fillval = 0.0  # Depth value for land cells
 
-        if version_control_dir is None:
-            raise ValueError(
-                "version_control_dir cannot be None. Version control is required for Topo objects. Old Topo Files can be added through from_topo_file() or from_topo_version_control() classmethods."
-            )
+        if git:
```
alperaltuntas, line 60: The extent of this if block seems excessive. At the least, `self.grid_file_path` assignment shouldn't be guarded by `if git`. Please refactor and make sure only git-related statements are guarded by `if git`.

mom6_forge/topo_editor.py
alperaltuntas, line 11: If TopoEditor truly requires git=true, please add a check.

alperaltuntas (changes requested):

mom6_forge/topo_editor.py
manishvenu: I just went through a made sure the Topoeditor works without version control, I also changed everything to use masked_depth.

mom6_forge/topo.py
manishvenu: I removed save because it was kind of irrelevant. It wasn't doing anything extra. The user can just call the write functions.

mom6_forge/topo.py
manishvenu: done

mom6_forge/topo.py
manishvenu: The grid file path is a git specific thing. If you don't have grid you don't create the version control directory at all.

manishvenu: I removed the part that wasn't needed in the if (the min depth initial command)

manishvenu:
Also, @alperaltuntas, If it saves you time, we can just get #88 reviewed and pushed and that'll work better.

mom6_forge/topo.py
alperaltuntas: > The grid file path is a git specific thing. If you don't have grid you don't create the version control directory at all.

Isn't `domain_dir` also a git specific thing? It's set to None under else, but `grid_file_path` is not. Why different treatment?

mom6_forge/topo_editor.py
alperaltuntas: Why have you changed everything to use masked_depth?

manishvenu: That's something that should have been changed earlier, I think we want the depth field to show the finished depth, so that the mask and depth don't look crazy different.

mom6_forge/topo.py
manishvenu: This is true, I removed that. I kept tcm is None just to be explicit it is not being set.

manishvenu:
Looks like the tests were always failing! It's passing now, and will merge!

PR #69, opened 2026-04-17
Create a new mask/depth calculation

manishvenu:
Details:
We knew this was going to be tricky, so here we go!
Support a new manual mask variable to preserve intermediate data. (This is because many functions raise ocean depth to above min depth or flattens it if you need ocean or land respectively, which loses the original data)

Input on variable names would be great!!! Not sure if I like _depth_raw as opposed to something else.

Changes:
1. There is no longer a _depth array, it is now a _depth_raw array
2. There is a new manual_mask variable (binary land/ocean)
3. The depth _property_ has changed! Now it computes the depth from the mask and the unmasked depth. Any ocean points on the mask must be > min_depth, if not, they are bumped to min_depth + 0.1, Any land points on the mask must be < min_depth, and are bumped to the self._land_fillval if not (which is enforced to be lower than min depth).
4. There is a new mask property that uses the depth_raw_mask if there is no manual_mask

Important Info:
Most, if not ALL, functions should use the .depth property and .mask property and associated edit commands!! These have all the necessary gaurdrails. The only functions that should impact the raw depth are the depth editing functions.  raw depth and manual mask are low-level topo object variables that should only be touched if needed!!


.docstr/bathymetry/index.rst
```diff
@@ -0,0 +1,107 @@
...
+**Version Control Integration**
+
+All bathymetry modifications are tracked through the TopoCommandManager (TCM):
+
+- Every edit creates a record (index, old_value, new_value)
+- Full undo/redo capability
+- Version history saved to git
+- Can restore any previous bathymetry state
+
+**Minimal Depth Enforcement**
```
alperaltuntas: Minimal -> Minimum

.docstr/bathymetry/index.rst
```diff
@@ -0,0 +1,107 @@
...
+
+The ``min_depth`` parameter ensures all ocean cells meet a minimum depth for model stability:
+
+- Ocean cells shallower than ``min_depth`` are boosted to ``min_depth + 0.1``
```
alperaltuntas, line 48: Can't we just set them to min_depth?

.docstr/bathymetry/index.rst
```diff
@@ -0,0 +1,107 @@
...
+    # Undo if needed
+    topo.tcm.undo()
+    
+    # Save changes
+    topo.save()
+
+Documentation Index
+-------------------
+
+- **Mask and Depth Calculation** — Details on the separation-of-concerns design for depth and masking
```
alperaltuntas, line 101: Not sure if `seperation-of-concerns` is the right term here. They are quite intertwined.

mom6_forge/topo.py
```diff
@@ -131,37 +137,69 @@ def from_topo_file(
...
+        When mask is set, returns masked depth: ocean cells ≥ min_depth+0.1, land cells set to land_fillval.
+        When mask is None, returns raw depth.
         """
-        return self._depth
+        if self._manual_mask is None:
+            return self._depth_raw
+
+        # Apply masking: ocean cells stay as-is (or bump ≥ min_depth+0.1), land cells → land_fillval
+        masked_depth = self._depth_raw.copy()
+        ocean_mask = self._manual_mask > 0  # 1 → True (ocean), 0 → False (land)
```
alperaltuntas: Why is `ocean_mask` necessary here? Isn't manual_mask already 0 or 1?

mom6_forge/topo.py
```diff
@@ -131,37 +137,69 @@ def from_topo_file(
...
         MOM6 grid depth array (m). Positive below MSL.
+
+        When mask is set, returns masked depth: ocean cells ≥ min_depth+0.1, land cells set to land_fillval.
+        When mask is None, returns raw depth.
         """
-        return self._depth
+        if self._manual_mask is None:
+            return self._depth_raw
+
+        # Apply masking: ocean cells stay as-is (or bump ≥ min_depth+0.1), land cells → land_fillval
```
alperaltuntas: One small thing: could you stick to plain ASCII in the code and comments? AI tools love sprinkling in characters like `≥`,`→`, etc, but they're awkward to type and search for. Perhaps we should add `CLAUDE.md` for things like this.

mom6_forge/topo.py
```diff
@@ -131,37 +137,69 @@ def from_topo_file(
...
+        depth: np.array or xr.DataArray
             2-D Array of ocean depth (m).
         """
+        self.depth_raw = depth  # Use the depth_raw setter
 
-        if np.isscalar(depth):
-            self.set_flat(depth)
-            return
+    @property
+    def depth_raw(self):
```
alperaltuntas: When do we need this property be available outside this class?

mom6_forge/topo.py
```diff
@@ -131,37 +137,69 @@ def from_topo_file(
...
+        Raw depth array before masking is applied.
+        """
+        return self._depth_raw
 
-        assert depth.shape == (
-            self._grid.ny,
-            self._grid.nx,
-        ), "Incompatible depth array shape"
+    @depth_raw.setter
+    def depth_raw(self, depth):
```
alperaltuntas: Again, do we need outside clients to have access to a depth_raw setter?

mom6_forge/topo.py
```diff
@@ -131,37 +137,69 @@ def from_topo_file(
...
+            ocean_mask,
+            xr.where(
+                masked_depth > self._min_depth, masked_depth, self._min_depth + 0.1
+            ),
+            self._land_fillval,  # Land cells → land_fillval
+        )
+        return masked_depth
 
     @depth.setter
     def depth(self, depth):
```
alperaltuntas, line 190: There is an important setter asymmetry here: Because depth.setter writes to _depth_raw but depth.getter applies masking, topo.depth = x; topo.depth won't round-trip when a mask is active. This should be documented.

alperaltuntas (changes requested):
Thanks @manishvenu. I left some comments in the code, but more generally, I'd like to propose a structural simplification on the mask side before we go deeper on details.

Replace _manual_mask / _depth_raw_mask with user_mask:
                                                                                                                                          
  Instead of `_manual_mask`, `_depth_raw_mask`, and `mask` properties/members, introduce a single `user_mask` / `_user_mask` property. And keep using `tmask`/`umask`/`vmask` as the external mask view.        
                  
When user_mask  is `None`, tmask derives from depth and min_depth; when set, it overrides. This collapses the PR's three mask-ish concepts into one editable input (`user_mask`) plus the existing derived views. Also drop land_fillval, and set land cells to 0.0.

mom6_forge/topo.py

mom6_forge/topo.py

mom6_forge/topo.py

mom6_forge/topo.py

.docstr/bathymetry/index.rst
manishvenu: Done

.docstr/bathymetry/index.rst
manishvenu: I'm pretty sure that still gets masked out, it needs be a little above iirc.

.docstr/bathymetry/index.rst
manishvenu: Done

mom6_forge/topo.py
manishvenu: Done

manishvenu:
> Thanks @manishvenu. I left some comments in the code, but more generally, I'd like to propose a structural simplification on the mask side before we go deeper on details.
> 
> Replace _manual_mask / _depth_raw_mask with user_mask:
> 
> Instead of `_manual_mask`, `_depth_raw_mask`, and `mask` properties/members, introduce a single `user_mask` / `_user_mask` property. And keep using `tmask`/`umask`/`vmask` as the external mask view.
> 
> When user_mask is `None`, tmask derives from depth and min_depth; when set, it overrides. This collapses the PR's three mask-ish concepts into one editable input (`user_mask`) plus the existing derived views. Also drop land_fillval, and set land cells to 0.0.

Hey @alperaltuntas,

I think I've made all the changes required, including this big structure change, which moved the mask calculation to `tmask`. A couple things:

1. `land_fillval` is a remnant from `apply_land_frac`, I just moved it to the class instead. If we don't need it that's great.
2. You can no longer do editing of depth with `topo.depth[index] = new_val`. 
3. This edits make it so tmask decides between the raw depth mask and a user_mask (preferred). tmask is then used in the depth property.

manishvenu:
@alperaltuntas I have made the changes we discussed in this PR with a new masked_depth property, adding back land_fillval and having the depth property just return the _depth array (renamed the _depth_raw back to _depth)

docs/source/index.rst
```diff
@@ -16,7 +16,12 @@ to learn about the `mom6_forge` tool.
 
    installation
    quickstart
+<<<<<<< HEAD:.docstr/index.rst
```
alperaltuntas: resolve this conflict please.

mom6_forge/edit_command.py
```diff
@@ -167,3 +241,64 @@ def reverse_deserialize(cls, data):
...
+            all_indices = list(np.ndindex(mask_array.shape))
+            new_values = mask_array.ravel().tolist()
+            old_values = [0] * len(
+                new_values
+            )  # placeholder, will be overwritten on __call__
+            return MaskEditCommand(
+                topo,
+                all_indices,
+                new_values,
+                old_values=old_values,
```
alperaltuntas, line 296: Why don't we set `old_values=None` here, instead of the placeholder?

alperaltuntas (approved):

PR #71, opened 2026-04-17
Add a small source bathy class

manishvenu:
This class is to reduce the amount of arguments that go into topo + raw dataset functions. It replaces config_dataset


mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,162 @@
...
+        if self.positive_down:
+            return self._da.values.astype(float)
+        else:
+            return -self._da.values.astype(float)
+
+    @property
+    def xesmf_ready_ds(self):
+        """Dataset with standardized coordinate names and positive-down elevation."""
+        ds = self.ds.copy()
+        ds.rename(
```
alperaltuntas: Shouldn't this be `ds = ds.rename(...`

or better yet, `ds = self.ds.rename(...`

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,162 @@
...
+        ds.depth.attrs["units"] = "meters"
+        ds.depth.attrs["standard_name"] = "height_above_reference_ellipsoid"
+        ds.depth.attrs["long_name"] = "Depth relative to sea level"
+        ds.depth.attrs["coordinates"] = "lon lat"
+        return ds
+
+    @property
+    def da(self):
+        """Raw elevation DataArray with source coordinate names (positive-up)."""
+        if not self.positive_down:
```
alperaltuntas: Judging by the docstring, should this be: `if self._positive_down:`

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,162 @@
...
+            return -self._da
+        else:
+            return self._da
+
+    @property
+    def ds(self):
+        """Raw dataset with source coordinate names (positive-up)."""
+        if not self.positive_down:
```
alperaltuntas: Again, I am confused here.

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,162 @@
...
+                ds,
+                np.array(lon_extent) + np.array([-buf, buf]),
+                self.lon_name,
+            )
+        else:
+            ds = ds.sel(
+                {self.lon_name: slice(lon_extent[0] - buf, lon_extent[1] + buf)}
+            )
+        self._ds = ds
+        self._da = ds[self.elevation_name].load()
```
alperaltuntas: Do we assume that elevation read from the file is positive-up? If so, we should have an assertion to confirm that assumption.

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,162 @@
...
+        lat_name="lat",
+        elevation_name="elevation",
+        positive_down=False,
+        buf=0.5,
+    ):
+        self.path = Path(path)
+        self.lon_name = lon_name
+        self.lat_name = lat_name
+        self.elevation_name = elevation_name
+        self.positive_down = positive_down  # depth should be positive down (ocean > 0) if True, otherwise positive up (ocean < 0)
```
alperaltuntas: I think supporting both positive_up and _down complicates things. Can we instead just pick one and enforce it throughout?

mom6_forge/_source_bathy.py
manishvenu: I was doing this not in-place, so I copied it, but to be honest, there's no reason not to have it xesmf ready.

mom6_forge/_source_bathy.py
manishvenu: Fixed, I'm removing this in favor of a one time edit.

mom6_forge/_source_bathy.py
manishvenu: Same here.

mom6_forge/_source_bathy.py
manishvenu: Same here.

mom6_forge/_source_bathy.py
manishvenu: Done.

mom6_forge/_source_bathy.py
manishvenu: So now there's only the one ds that is already xesmf ready. (new rename func)

mom6_forge/_source_bathy.py
manishvenu: There is a func that makes sure depth is positive.

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,147 @@
...
+    elevation_name : str — elevation variable (positive-up). Default ``"elevation"``.
+    """
+
+    def __init__(
+        self,
+        topo,
+        path,
+        lon_name="lon",
+        lat_name="lat",
+        depth_name="elevation",
```
alperaltuntas: To me, the variable names "depth" and "elevation" contradictory. (I assume that the convention is *depth* is positive below MSL while *elevation* is the opposite.) So `depth_name="elevation"` is confusing me. Can we stick to one or the other uniformly within this class? The docstring above is also inconsistent with the actual parameter name. My preference is to use "depth" throughout. But if there is a good reason to use "elevation", then we should do so consistently within this class.

mom6_forge/topo.py
```diff
@@ -442,6 +435,36 @@ def supergridmask(self):
...
+        latitude_coordinate_name,
+        vertical_coordinate_name,
+        positive_down=False,
+        buf=0.5,
+    ):
+        """Set a :class:`SourceBathy`, creating and slicing a new one
+        only when the path or coordinate names differ from the current cache."""
+        path = Path(bathymetry_path)
+        if (
+            self.src is None
```
alperaltuntas: So, if src is already set, this method will silently disregard the new source being provided, without any indication to the user? If we don't want the user the override the src (and if so, why?), then this method should throw an error when the user attempts to do so.

mom6_forge/_source_bathy.py
alperaltuntas: Could you elaborate on how you fixed this?

mom6_forge/topo.py
manishvenu: WE're letting the user ovverride src now!

mom6_forge/_source_bathy.py
manishvenu: Done

manishvenu:
This next commit makes the longitude_slicer work for regional domains & replaces config_dataset with src_bathy funcs (which is what this class is supposed to do).

Because this is a divergence from regional_mom6, I changed the version to 0.1.0

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,147 @@
...
+    cached here so repeated calls with the same source file skip the
+    expensive sub-sampling step.
+
+    Parameters
+    ----------
+    path : str or Path
+    lon_name : str   — longitude coordinate name. Default ``"lon"``.
+    lat_name : str   — latitude coordinate name. Default ``"lat"``.
+    depth_name : str — depth variable. Default ``"depth"``.
+    """
```
alperaltuntas, line 47: Please document all of the parameters.

mom6_forge/_source_bathy.py
alperaltuntas: I still see the term elevation in the docstrings and in testing.

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,147 @@
...
+        """2-D depth array, depth positive (ocean > 0)"""
+        return self.ds[self.depth_name].values
+
+    @property
+    def ds(self):
+        """Raw dataset with source coordinate names (positive-down)."""
+        return self._ds
+
+    def __repr__(self):
+        shape = self._da.shape if self._da is not None else "not loaded"
```
alperaltuntas: I can't find `self._da` or `self.elevation_name` defined anywhere else in the class. Am I missing something, or is this a bug? If it is a bug, what concerns me more is that the tests didn't reveal it. Please make sure all new methods are tested.

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,147 @@
+"""Source bathymetry loader for mom6_forge.
+
+``SourceBathy`` is a lightweight data container for a regional slice of a
+source bathymetry dataset.  Users who call pipeline
+methods directly
+should construct a ``SourceBathy`` explicitly::
+
+    from mom6_forge._source_bathy import SourceBathy
+    src = SourceBathy("gebco_2023.nc").slice_to_domain(topo)
```
alperaltuntas: Is this example still valid? Please review all doctrings. There seems to be many stale/inaccurate documentation.

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,147 @@
...
+from pathlib import Path
+from mom6_forge.utils import longitude_slicer
+
+
+class SourceBathy:
+    """Regional slice of a source bathymetry dataset (e.g. GEBCO).
+
+    Holds the loaded, domain-clipped, ESMF prepped elevation DataArray together with its
+    coordinate-name metadata.  Per-cell depth statistics are computed and
+    cached here so repeated calls with the same source file skip the
```
alperaltuntas: Could you describe how caching is implemented?

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+
+    def __init__(
+        self,
+        topo,
+        path,
+        lon_name="lon",
+        lat_name="lat",
+        depth_name="depth",
+        is_input_positive_below_msl=True,
```
alperaltuntas, line 56: The default for this arg is `True` here, but is `False` everywhere else. Is this discrepancy intentional?

mom6_forge/topo.py
alperaltuntas, line 942: We should stick to numpy docstring format. (This was introduced prior to this PR, but just mentioning here as I've just noticed it.)

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+        self.lon_name = "lon"
+        self.lat_name = "lat"
+        self.depth_name = "depth"
+        self._ds.depth.attrs["missing_value"] = (
+            -1e20
+        )  # missing value expected by FRE tools
+        self._ds.depth.attrs["_FillValue"] = -1e20
+        self._ds.depth.attrs["units"] = "meters"
+        self._ds.depth.attrs["standard_name"] = "height_above_reference_ellipsoid"
```
alperaltuntas: Please update these attributes.

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,147 @@
...
+        if "units" not in self._ds[self.lon_name].attrs:
+            self._ds[self.lon_name].attrs["units"] = "degrees_east"
+        if "units" not in self._ds[self.lat_name].attrs:
+            self._ds[self.lat_name].attrs["units"] = "degrees_north"
+
+    def _slice_to_domain(self, topo, buf=0.5):
+        """Load and clip elevation to the topo grid extent plus ``buf`` degrees.
+
+        Handles the global-longitude seam automatically.  Mutates ``self``
+        in place and returns ``self`` for chaining.
```
alperaltuntas: please update/correct this line and all other stale/incorrect docstrings.

mom6_forge/_source_bathy.py
manishvenu: It's out of date, I'll update it & add a test

mom6_forge/_source_bathy.py
manishvenu: Done

mom6_forge/_source_bathy.py
manishvenu: Done

mom6_forge/_source_bathy.py
manishvenu: That is implemented in the next PR! I changed it to "stored"

mom6_forge/_source_bathy.py
manishvenu: GEBCO, the product we mainly use, is elevation based so you always need to reverse the sign. If we're calling it "depth_name" I think we should have default is_input_positive_below_msl is True, which means everywhere else need to flip the sign if we use a sample gebco dataset

mom6_forge/topo.py
manishvenu: Ok

mom6_forge/_source_bathy.py
manishvenu: Could you explain what you mean? I removed the standard & long name because they may not be true (I was just carrying over regional-mom6 stuff)

mom6_forge/_source_bathy.py
manishvenu: Ok

mom6_forge/topo.py
```diff
@@ -52,6 +53,7 @@ def __init__(self, grid, min_depth, version_control_dir="TopoLibrary", git=True)
             None  # Binary ocean/land mask (None = no mask applied)
         )
         self._min_depth = min_depth
+        self._src = None  # cached SourceBathy; set by _set_src()
```
alperaltuntas: I still see references to caching. Are these still relevant?

mom6_forge/topo.py
```diff
@@ -478,6 +471,28 @@ def apply_edit(self, cmd, skip_version_control=False, cmd_type=CommandType.COMMA
...
+        self,
+        bathymetry_path,
+        longitude_coordinate_name,
+        latitude_coordinate_name,
+        vertical_coordinate_name,
+        is_input_positive_below_msl=False,
+        buf=0.5,
+    ):
+        """Set a :class:`SourceBathy`, creating and slicing a new one
+        only when the path or coordinate names differ from the current cache."""
```
alperaltuntas: I still see references to caching. Are these still relevant?

mom6_forge/topo.py
```diff
@@ -1066,7 +944,7 @@ def tidy_dataset(
             fill_channels (Optional[bool]): Whether to fill in diagonal channels.
                 This removes more narrow inlets, but can also connect extra islands to land.
                 Default: ``False``.
-            positive_down (Optional[bool]): If ``False`` (default), assume that
+            is_input_positive_below_msl (Optional[bool]): If ``False`` (default), assume that
```
alperaltuntas: please update docstrings.

mom6_forge/_source_bathy.py
```diff
@@ -0,0 +1,345 @@
...
+                new_data = new_data.assign_coords({lon: new_lon})
+
+                ## Choose the number of lon points to take from the middle, including a buffer.
+                ## Use this to index the new global dataset
+                num_lonpoints = int(
+                    int(data[lon].shape[0] * (central_longitude - longitude_extent[0]))
+                    // lon_span
+                )
+
+        data = new_data.isel(
```
alperaltuntas, line 345: new_data and num_lonpoints are initialized only if the above if conditional is correct, but you use them here regardless.

mom6_forge/topo.py
manishvenu: No, removed

mom6_forge/topo.py
manishvenu: No

mom6_forge/topo.py
manishvenu: Done.

mom6_forge/_source_bathy.py
manishvenu: Yeah it should error out. I'll fix it.

alperaltuntas (approved):

PR #72, opened 2026-04-17
Compute Topo Stats (Frank Method)

manishvenu:
Details:
A new method to create grid cell stats, converted from frank's method. I only changed chlorophyll to use this (it was incorrectly not doing subsampling at all)


manishvenu:
@alperaltuntas This PR may require more intense review, it's one of two main functions in the bathy workflow, and I'm hoping it'll work with global grids.

mom6_forge/mapping.py
```diff
@@ -994,6 +994,148 @@ def regrid_dataset_via_xesmf(
...
+        Corner (q-point) coordinates of the low-res grid.
+    nx_sub, ny_sub : int
+        Number of sub-points per cell in each direction.
+
+    Returns
+    -------
+    sub_lon, sub_lat : np.ndarray  shape (ny, nx, ny_sub, nx_sub)
+    """
+    assert type(qlon) == type(qlat) == np.ndarray, "qlon and qlat must be numpy arrays"
+
```
alperaltuntas, line 1015: Does this function support all `ny_sub` and `nx_sub`, or do `nx` and `dx` need to be divisible by them?

mom6_forge/mapping.py
```diff
@@ -994,6 +994,148 @@ def regrid_dataset_via_xesmf(
...
+    qlon, qlat : np.ndarray  shape (ny+1, nx+1)
+        Corner (q-point) coordinates of the low-res grid.
+    nx_sub, ny_sub : int
+        Number of sub-points per cell in each direction.
+
+    Returns
+    -------
+    sub_lon, sub_lat : np.ndarray  shape (ny, nx, ny_sub, nx_sub)
+    """
+    assert type(qlon) == type(qlat) == np.ndarray, "qlon and qlat must be numpy arrays"
```
alperaltuntas: Pleaes use `isinstance`.

mom6_forge/topo.py
```diff
@@ -747,6 +747,90 @@ def set_bowl(self, max_depth, dedge, rad_earth=6.378e6, expdecay=400000.0):
...
+        Returns
+        -------
+        xr.Dataset  —  ``OCN_FRAC``, ``D_mean``, ``D_min``, ``D_max``, ``D2_mean``.
+        """
+        assert (
+            self.src is not None
+        ), "Source bathymetry must be loaded to compute topo stats"
+        src = self.src
+        if hasattr(self, "_stats") and self._stats is not None:
+            return self._stats
```
alperaltuntas: Once _stats is computed, this method will always return that stats regardless of the arguments to this method, which is wrong.

mom6_forge/mapping.py
```diff
@@ -994,6 +994,148 @@ def regrid_dataset_via_xesmf(
...
+
+
+def regrid_with_subsampling(
+    input_dataset,
+    qlon,
+    qlat,
+    nx_sub,
+    ny_sub,
+    regridding_method="nearest_s2d",
+):
```
alperaltuntas, line 1075: Judging by the reshaping logic below, this function works only for 2D `input_datasets`. If that's indeed the case, please add an assertion that checks that constraint.

alperaltuntas (changes requested):

mom6_forge/mapping.py
manishvenu: Done

mom6_forge/mapping.py
manishvenu: No? It supports all nx_sub ny_sub b/c you're dividing one grid cell by those sub values. not nx or dx.

mom6_forge/topo.py

mom6_forge/mapping.py
manishvenu: done.

mom6_forge/topo.py
```diff
@@ -775,6 +775,100 @@ def set_bowl(self, max_depth, dedge, rad_earth=6.378e6, expdecay=400000.0):
...
+
+        depth_ocean = np.where(is_ocean, depth_sub, np.nan)
+        with np.errstate(all="ignore"):
+            D_mean = np.nanmean(depth_ocean, axis=(-2, -1))
+            D_min = np.nanmin(depth_ocean, axis=(-2, -1))
+            D_max = np.nanmax(depth_ocean, axis=(-2, -1))
+            D2_mean = np.nanmean(depth_ocean**2, axis=(-2, -1))
+
+        dims = ["ny", "nx"]
+        self.src._stats = xr.Dataset(
```
alperaltuntas: SourceBathy does not declare a _stats attribute. This line both reaches into a private member of another class and creates it from the outside (which is worse than a normal private access, something we frequently do, but really shoudln't). Please either keep the cache on Topo (a dict keyed by (id(src), nx_sub, ny_sub, mask_hmin)), or alternatively declare self._stats = None in SourceBathy.__init__ and introduce a property getter and setter.

alperaltuntas (changes requested):

mom6_forge/topo.py
manishvenu: I did the Source Bathy with a getter/setter with another getter on Topo to also get the stats.

mom6_forge/chl.py
```diff
@@ -88,21 +89,27 @@ def interpolate_and_fill_seawifs(
...
+                "chlor_a": xr.DataArray(
+                    src_data[t, ::-1, :].values,
+                    dims=["lat", "lon"],
+                    coords={"lat": src_lat, "lon": src_lon},
+                )
+            }
+        )
+
+        # Regrid to super-sampled sub-point grid and average back to model grid
+        q_sub = regrid_with_subsampling(
```
alperaltuntas: Instead of computing regridding weight at every timestep, can we compute once, and reuse in subsequent time entries?

alperaltuntas (changes requested):

mom6_forge/chl.py
manishvenu: Done! I added it in regrid_dataset_via_xesmf if you pass in a saved weights path. (It's not default saved because saved weights can be sneakily used in the regridder and it's sometimes unclear why it fails)

alperaltuntas: Could you elaborate on "(It's not default saved because saved weights can be sneakily used in the regridder and it's sometimes unclear why it fails)"

manishvenu: Yes, if you save weights from a regridding under "weights.nc", then have a different regridding operation with a different grid and try to reuse the weights "weights.nc" - which can happen when you have automatic workflows like regional-mom6 where the user doesn't know everything that goes on under the hood - The error you get isn't clear that using a wrongly shaped weights file is the error. that's the problem.

alperaltuntas: Yeah, that would be a terrible way to reuse weights. Can you consider better approaches? Like, how about building the xe.Regridder once in memory outside the time loop and call it per timestep.

manishvenu: I like this approach becaue the user can specify a specific weights path to use. If they don't want that (which is default) it won't reuse that weights.

I'm also happy to build the weights before the chl timeloops but this is a bit more deeper level, which I prefer? But if you would rather have it done explictly that's fine to.

alperaltuntas: My original concern was about performance. The original implementation recomputes the same weights every timestep. In your current implementation, the default behavior is still the same. That is, the same weights are computed for every time step.

manishvenu: Because the weights file is specified regrid_dataset_via_xesmf will check if they already exist and reuse that weights file instead of recomputing. That's a feature on xesmf.Regridder with reuse_weights = True

manishvenu: The testing for regrid_with_subsampling runs twice, once with the saved weights -> and we know it's using those weights becaue I changed the regridding method and it still gave the same result.

alperaltuntas: Yes, reuse_weights=True skips the weights computation. But every timestep still opens and reads the file, deserializes the mapping weight matrix, and creates a new xe.Regridder object. So the regridder is built N times when we only need 1. Not to mention all the new complexity in the APIs.

manishvenu: Ok that's true, I'll see if I can build the Regridder outside of the loop first.

manishvenu: Turns out it was easy to get rid of the time loop for regridding. I guess this discussion is tabled for now but will show up again for #81

alperaltuntas: With this new approach of remapping all time steps at once, how likely is it to exhaust the memory? Have oyou done profiling?

manishvenu: I ran the derecho-only test in test_chl and it works fine -> it's 12 time steps (it's one a month for the dataset we provide with CrocoDash, SeaWifs)

alperaltuntas: Do we expect the users to process more than 12 timesteps? Do we expect the users to remap from/to grids that are higher resolution than what you tested? A comparison of memory consumption of previous and current version would help how likely this would break in production.

manishvenu: I don't expect people to do more, but it's definetely possible. I'll see if I can generate a dataset to test that to a high resolution grid & compare to the previous implementation.

alperaltuntas: A more robust way would be to still map per timestep, but to generate the remapping weights only once, as I originally suggested.

manishvenu: Ok, since the memory is the issue, I'll write it that way.

manishvenu: Ok the regridder is saved in the time loop

alperaltuntas (approved):

PR #73, opened 2026-04-17
Mask function from Compute Stats

manishvenu:
Add a mask function from compute stats


mom6_forge/topo.py
```diff
@@ -873,6 +873,53 @@ def _compute_stats(self, nx_sub, ny_sub, mask_hmin):
...
+        bilinear interpolation of the Q-point corners and snaps each to the
+        nearest source pixel. A cell is ocean if its ocean sub-point fraction
+        (OCN_FRAC) meets or exceeds mask_threshold.
+
+        Per-cell depth statistics (D_mean, D_min, D_max, D2_mean) are on
+        the source bathymetry object for downstream use by this function and others.
+
+        Parameters
+        ----------
+        self.src.stats : SourceBathy stats
```
alperaltuntas: I don't see this parameter in the method signature.

mom6_forge/topo.py
```diff
@@ -873,6 +873,53 @@ def _compute_stats(self, nx_sub, ny_sub, mask_hmin):
...
+
+        Returns
+        -------
+        xr.DataArray
+            Binary ocean mask on the T-grid (1 = ocean, 0 = land),
+            dims ``["ny", "nx"]``.
+        """
+
+        assert (
+            self.src.stats is not None
```
alperaltuntas, line 911: Please add a test that triggers a violation of this assertion.

mom6_forge/topo.py
```diff
@@ -873,6 +873,53 @@ def _compute_stats(self, nx_sub, ny_sub, mask_hmin):
...
 
+    def generate_mask_from_stats_ocean_frac(
+        self,
+        mask_threshold=0.5,
+    ):
+        """
+        Generate an ocean mask by uniform sub-sampling of the source
+        bathymetry. Mirrors the algorithm in tx2_3's create_model_topo.f90.
+
+        _compute_stats must be called first
```
alperaltuntas: This is a public method that expects a client to call a private method, which is awkward. Can we call _compute_stats from inside this method.

alperaltuntas (changes requested):

mom6_forge/topo.py
manishvenu: Hey Alper, 

Before I implement this, one of the things I wanted to do with these methods is have them not have a super long parameter list. Would it be okay to hae compute_stats be public? I don't see any reason it can't be.

mom6_forge/topo.py

mom6_forge/topo.py
manishvenu: Done

mom6_forge/topo.py
alperaltuntas: sounds good.

manishvenu: Ok it is finished!

alperaltuntas (approved):

PR #76, opened 2026-04-20
Mask into Topo Editor + Rectangle Selector

manishvenu:
CLOSED IN FAVOR OF #83

Changes:

1. Add the mask variable to the topo editor in the form of using the topo mask property and adding a mask edit field that can be switched from land to ocean
2. Add a toggle to do rectangle select, to set large chunks of bathymetry to a specific value
3. Minor cleaning


manishvenu:
CLOSED IN FAVOR OF #83

PR #77, opened 2026-04-21
Display Stats in the Topo Editor on zoom

manishvenu:
REPLACED BY #84

Changes:

1. Once the zoom reaches a threshold (40 cells) show the cell stats in the cell 
2. Create new buttons to set the depth to those stat numbers if wanted.
Question:
You can use the buttons even if you don't see the stats, which only show up when you zoom in. Is that good behavior?


manishvenu:
REPLACED BY https://github.com/NCAR/mom6_forge/pull/84

PR #81, opened 2026-04-27
Cressman Depth

manishvenu:
Replaces #80

Changes (Mapping.py):

1. `compute_cressman_weights' - the actual math calculation to calculate weights using the cressman formula, returns in the form of xesmf weights like the other mapping function `write_mapping_file` but without knowing the ESMF mesh (some fields are not filled)
2. `regrid_dataset_via_cressman` - this function is just the wrapper that calls `compute_cressman_weights` then `regrid_dataset_via_xesmf` to compute the new depth using the cressman weights
3. `source_to_dst` and `dst_to_source` - functions that take a weights dataset and a 2d index and return the useful information like what source points are weighted for a specific destination point.

Changes (utils.py + topo.py):

1. `iterative_fill`. Using a set number of iters, you basically just fill in an unfilled cell from its numbers. Every iter fills in a new neighbor. So an iteration of 100, fills 100 cells over. 
2. `direct_cressman_interp` - It's the topo.py wrapper for cressman. It calls `regrid_dataset_via_cressman` and then `iterative_fill` with some minimal necessary formatting.

Changes (other):

1. A new notebook to explain what cressman is.
3. tests


mom6_forge/topo.py
```diff
@@ -1018,6 +1022,79 @@ def mpi_set_from_dataset(
             "Configuration complete. Ready for regridding with MPI. See documentation for more details."
         )
 
+    def direct_cressman_interp(
```
alperaltuntas, line 1138: Please add a test that covers this method.

alperaltuntas (changes requested):

mom6_forge/topo.py
manishvenu: Done!

mom6_forge/topo.py
```diff
@@ -1018,6 +1022,69 @@ def mpi_set_from_dataset(
...
+
+            w = \\left(\\frac{L^2 - r^2}{L^2 + r^2}\\right)^{c}
+
+        where ``r`` is the great-circle arc distance and ``c = cressman_exp``.
+        Only source points with positive depth (ocean) contribute, so depth
+        estimates are never contaminated by land elevations.
+
+        Weights are computed in :func:`~mom6_forge.mapping.compute_cressman_weights`,
+        saved to an ESMF-compatible netCDF via
+        :func:`~mom6_forge.mapping.write_cressman_weights`, and applied through
```
alperaltuntas: Please update stale docstrings.

mom6_forge/mapping.py
```diff
@@ -994,6 +1020,421 @@ def regrid_dataset_via_xesmf(
...
+
+def regrid_dataset_via_cressman(
+    src_ds: xr.Dataset,
+    dst_ds: xr.Dataset,
+    weights_path: Path,
+    smooth_scl: float = 2.0,
+    cressman_exp: float = 2.0,
+    write_to_file: bool = False,
+    output_path: Path = Path("cressman_regridded.nc"),
+) -> tuple[np.ndarray, np.ndarray]:
```
alperaltuntas: Please correct the return type here and in the docstring.

alperaltuntas (changes requested):
Final, minor comments.

mom6_forge/topo.py
manishvenu: Fixed (tho this one looks right)

mom6_forge/mapping.py
manishvenu: Fixed

alperaltuntas (approved):

PR #83, opened 2026-04-30
Masking in Topo Editor + Rectangle Selector

### LINKED ISSUE #134: Change large areas in topography editor
I have a use case where I wanted to select a large area of ocean (using the topo editor) and change it to land. When editing the ocean depth in the topo editor we currently only have the option of selecting single grid cells at a time. It would be good to have the ability to change the depth off all grid cells in a user-defined area at the same time.

manishvenu:
Replaces #76 , Closes https://github.com/CROCODILE-CESM/CrocoDash/issues/134

Changes:

Add the mask variable to the topo editor in the form of using the topo mask property and adding a mask edit field that can be switched from land to ocean
Add a toggle to do rectangle select, to set large chunks of bathymetry to a specific value
Minor cleaning


mom6_forge/topo_editor.py
```diff
@@ -19,16 +23,27 @@ def __init__(self, topo, build_ui=True):
...
+        self.trigger_refresh()
 
         # --- Initialize the widget layout ---
         super().__init__([self._control_panel, self._interactive_plot])
 
+    @property
+    def active_cells(self):
+        if hasattr(self, "_selected_cells") and self._selected_cells:
+            return list(self._selected_cells)
+        elif self._selected_cell is not None:
```
alperaltuntas: This elif statement doesn't make sense to me. If the above if statement evaluates to true, then this elif branch will be skipped. If the if statement is false, then this elif branch will be false too, or will lead to a runtime error.

mom6_forge/topo_editor.py
```diff
@@ -431,11 +504,56 @@ def construct_observances(self):
         # Git/domain controls
         self._git_create_branch_button.on_click(self.on_git_create_branch)
         self._git_checkout_button.on_click(self.on_git_checkout)
-        self._display_mode_toggle.observe(
```
alperaltuntas, line 496: Why is this observe call no longer necessary?

alperaltuntas (changes requested):
I left a couple of comments, but more importantly, I am unable to test these changes because TopoEditor doesn't work in jupyter via the `mom6_forge` conda environment. I am getting:

```
ImportError: libpmi.so.0: cannot open shared object file: No such file or directory
```

when I attempt to run TopoEditor via the (newly installed) mom6_forge conda environment. We need to fix that issue first before I can evaluate this PR.

mom6_forge/topo_editor.py
manishvenu: Ahh that's maybe my bad. The different is an "s": selected_cells, versus selected_cell. Want me to change that to more different?

mom6_forge/topo_editor.py
manishvenu: It was repeated twice.

manishvenu:
@alperaltuntas, I am unable to reproduce your error directly, I tightened up the ipympl to be <0.10.0 and fixed a seperate error so that potentially may help.

alperaltuntas (changes requested):
Could you resolve the conflicts with the latest main? The version control conflicts will need extra attention.

manishvenu:
It's merged!

mom6_forge/topo_editor.py
```diff
@@ -430,6 +501,14 @@ def construct_observances(self):
             self.on_depth_change, names="value", type="change"
         )
 
+        # Mask change observer for selected cell
+        self._mask_specifier.observe(self.on_mask_change, names="value", type="change")
+        self._clear_user_mask_button.on_click(self.clear_user_mask)
+
+        # Mask change observer for selected cell
```
alperaltuntas: Duplicate block

tests/conftest.py
```diff
@@ -5,9 +5,22 @@
 import xarray as xr
 from mom6_forge.grid import Grid
 from mom6_forge.topo import Topo
+from mom6_forge.topo_editor import TopoEditor
 from mom6_forge._source_bathy import SourceBathy
 
 
+@pytest.fixture
+def get_rect_topo(tmp_path, get_rect_grid):
```
alperaltuntas: duplicate

mom6_forge/topo_editor.py
```diff
@@ -6,6 +8,8 @@
 from matplotlib.ticker import MaxNLocator
 from mom6_forge.edit_command import *
 from mom6_forge.git_utils import *
+from matplotlib.widgets import RectangleSelector
+from matplotlib.backends.backend_agg import FigureCanvasAgg
```
alperaltuntas: remove unused import

alperaltuntas (changes requested):

mom6_forge/topo_editor.py
manishvenu: Removed

tests/conftest.py
manishvenu: Removed

mom6_forge/topo_editor.py
manishvenu: Removed

alperaltuntas:
@manishvenu This new feature looks great! I have several UX suggestions:
 - While "Zoom to rectangle" tool is selected on the top right corner,  clicking "Rectangle select" appears to active rectangle selection, but "Zoom to rectangle" remains active instead. When the user selects "Rectangle select", the "Zoom to rectangle" should be unset.
 - For the "Rectangle Select" Button, I suggest the following new labels: 
   - When off: `Mark Rectangular Region`
   - When on: `Unmark Region`
 - When I attempt to change the depth of a selected rectangular region, it takes quite a long time to apply that change. (Maybe it does so for pointwise changes too). In fact, it broke my jupyterhub session after a few applications. Please find the bottleneck and improve efficiency. This may be due to the version control overhead. If that's the case, I'd say we disable it by default.

manishvenu:
hey @alperaltuntas, but disbale version control, are you thinking overall, have version control disabled by default, or just in the topo editor disable version control.

alperaltuntas:
> hey @alperaltuntas, but disbale version control, are you thinking overall, have version control disabled by default, or just in the topo editor disable version control.

Yes, I was thinking disabling version control by default, unless the user opts in. Does the following PR address this issue: https://github.com/NCAR/mom6_forge/pull/106

manishvenu:
> > hey @alperaltuntas, but disbale version control, are you thinking overall, have version control disabled by default, or just in the topo editor disable version control.
> 
> Yes, I was thinking disabling version control by default, unless the user opts in. Does the following PR address this issue: #106

Hey Alper,

Yes, it's supposed to. Do you want me to still disable? It's probably not a bad idea.

alperaltuntas:
> Do you want me to still disable? It's probably not a bad idea.

Yeah, I think that's the safer approach.

manishvenu:
@alperaltuntas It's disabled , I'll fix the CI issue, it's not showing up locally so it's just a little buggy.

mom6_forge/topo_editor.py
```diff
@@ -413,14 +472,24 @@ def _select_cell(self, i, j):
...
 
-        self._selected_cell = (i, j, polygon)
+        self._selected_cell = (j, i, polygon)
 
         # UI updates
         if hasattr(self, "_selected_cell_label"):
             self._selected_cell_label.value = f"Selected cell: {i}, {j}"
         if hasattr(self, "_depth_specifier"):
             self._depth_specifier.disabled = False
             self._depth_specifier.value = self.topo.masked_depth.data[j, i]
```
alperaltuntas: Since `on_depth_change` now reads `old_values` from raw `depth` (not `masked_depth`), this line creates an inconsistency: selecting a cell sets the field to the masked value, which fires `on_depth_change` with `new_val = masked` vs `old = raw`. For any land or shallow cell where `masked_depth` is not equal to `depth` (for example a land cell becomes fillval 0.0 while raw `_depth` is nonzero), the no-op guard fails and a spurious `DepthEditCommand` rewrites the raw depth and pushes an undo entry, just from clicking the cell.

Note that storing raw `old_values` is correct for undo, since `reverse_deserialize` restores `_depth` to `old_values`.

mom6_forge/topo_editor.py
alperaltuntas: This is no longer correct since _selected_cell return in j, i order. Please also check the remainder of the code.

alperaltuntas (changes requested):
In addition to the line comments I added, here are two more comments:
-  I believe the Mean/Max/Min buttons work for single selected cells only. If that's the case, they should be disabled when rectangle select is made.
- Toggling rectangle mode ON doesn't clear the prior single-cell selection.

mom6_forge/topo_editor.py
manishvenu: Ok so the fix for this would be to have the depth specifier show the actual raw depth. The plot still shows the masked depth, which I think is fine. I've done this fix.

mom6_forge/topo_editor.py
manishvenu: Fixed.

manishvenu:
@alperaltuntas I made the stats able to be used in rect mode but the label removed the selected cell stats. The prior selected cell selection is also now cleared

mom6_forge/topo_editor.py
```diff
@@ -162,6 +72,19 @@ def construct_control_panel(self):
         self._selected_cell_label = widgets.Label(
             "Selected cell: None (double click to select a cell)."
         )
+        self._rect_or_single_select_button = widgets.ToggleButton(
```
alperaltuntas: Can we instead use `ToggleButtons` here? Something like:

```
widgets.ToggleButtons(
    options=['Single Cell', 'Rectangular Area'],
    description='Selection Mode:',
    ...
)
```

alperaltuntas (changes requested):
One final comment. Otherwise, I think this is ready.

mom6_forge/topo_editor.py
manishvenu: Done

alperaltuntas:
Once the failing test is fixed, this can be merged.

manishvenu:
Hey @alperaltuntas, Can you approve? I can't merge until it is approved.

alperaltuntas (approved):

PR #84, opened 2026-04-30
Display Stats in the Topo Editor on zoom

manishvenu:
REPLACES #77 

Changes:

Once the zoom reaches a threshold (40 cells) show the cell stats in the cell
Create new buttons to set the depth to those stat numbers if wanted.
Question:
You can use the buttons even if you don't see the stats, which only show up when you zoom in. Is that good behavior?


mom6_forge/topo_editor.py
```diff
@@ -415,7 +566,9 @@ def construct_observances(self):
 
         # Double click event for cell selection on the plot
         self.fig.canvas.mpl_connect("button_press_event", self.on_double_click)
-
+        # Zoom-dependent stats overlay
+        self.ax.callbacks.connect("xlim_changed", self._on_zoom_change)
+        self.ax.callbacks.connect("ylim_changed", self._on_zoom_change)
```
alperaltuntas: A single zoom changes both xlim and ylim. With the ebaove two lines, a single zoom event will call _on_zoom_change twice.

mom6_forge/topo_editor.py
```diff
@@ -507,6 +665,37 @@ def on_depth_change(self, change):
...
+
+        if ds is None or stat_name not in ds.data_vars:
+            return None
+
+        return float(ds[stat_name].data[j, i])
+
+    def set_depth_to_mean(self, b):
+        """Set the selected cell's depth to the mean value."""
+        val = self._get_statistic_value("D_mean")
+        if val is not None:
```
alperaltuntas: replace the if statement here as: `if val is not None and np.isfinite(val)`

mom6_forge/topo_editor.py
```diff
@@ -507,6 +665,37 @@ def on_depth_change(self, change):
...
+            self._depth_specifier.value = val
+
+    def set_depth_to_max(self, b):
+        """Set the selected cell's depth to the max value."""
+        val = self._get_statistic_value("D_max")
+        if val is not None:
```
alperaltuntas: replace the if statement here as: `if val is not None and np.isfinite(val)`

mom6_forge/topo_editor.py
```diff
@@ -507,6 +665,37 @@ def on_depth_change(self, change):
...
+            self._depth_specifier.value = val
+
+    def set_depth_to_min(self, b):
+        """Set the selected cell's depth to the min value."""
+        val = self._get_statistic_value("D_min")
+        if val is not None:
```
alperaltuntas: replace the if statement here as: `if val is not None and np.isfinite(val)`

mom6_forge/topo_editor.py
```diff
@@ -343,6 +372,118 @@ def trigger_refresh(self):
...
+            & (qlat[1:, :-1] >= ylim[0])
+            & (qlat[1:, :-1] <= ylim[1])
+            & (qlat[1:, 1:] >= ylim[0])
+            & (qlat[1:, 1:] <= ylim[1])
+        )
+
+        # Always clear old annotations first
+        if hasattr(self, "_cell_stat_texts"):
+            for t in self._cell_stat_texts:
+                t.remove()
```
alperaltuntas: This loop calls `t.remove()` on each old text label. If a label is already detached (for example after the axes are cleared), remove() raises. Because this runs inside the zoom callback and  the error happens before self._cell_stat_texts = [], the list never gets reset, so every later zoom fails again and the overlay stops working for the rest of the session. The selection-polygon code in `_select_cell` already wraps removal in try/except. Could we do the same here?

mom6_forge/topo_editor.py
```diff
@@ -343,6 +372,118 @@ def trigger_refresh(self):
         self._min_depth_specifier.value = self.topo.min_depth
         self.update_undo_redo_buttons()
 
+    def _draw_cell_stats(self, visible):
+        """Draw per-cell stat annotations directly on each visible cell."""
+        # Clear previous annotations
+        if hasattr(self, "_cell_stat_texts"):
+            for t in self._cell_stat_texts:
```
alperaltuntas: This method is only called from `_on_zoom_change`, which clears `_cell_stat_texts` and resets it to [] immediately before calling here. So this loop always runs on an empty list and never removes anything. If this block was added in anticipation of `_draw_cell_stats` being called from elsewhere in the future, then this block (`# Clear previous annotations`) should be made  a shared method to avoid code duplication. (Note, I have a fix suggestion for the corresponding `_on_zoom_change` block, which would need to be applied here too.)

alperaltuntas (changes requested):
In addition to the below comments, please merge the latest main. Thanks!

mom6_forge/topo_editor.py
manishvenu: Done

mom6_forge/topo_editor.py
manishvenu: Done

mom6_forge/topo_editor.py
manishvenu: Done

mom6_forge/topo_editor.py
manishvenu: Done

mom6_forge/topo_editor.py
manishvenu: Done

mom6_forge/topo_editor.py
manishvenu: I'll remove this part for now.

manishvenu:
@alperaltuntas,

Claude debugged this as when home is clicked, it calls the set_xlim and set_ylim which rewrites the stats text twice


```
The matplotlib toolbar home button resets the view by calling set_xlim(original) and set_ylim(original) as two separate, sequential calls.

The callback was only connected to xlim_changed. So here's what happened when you hit home after zooming in:

set_xlim(original) fires → xlim_changed → _on_zoom_change runs
Inside _on_zoom_change, it reads the current xlim (now restored) and ylim (still zoomed)
With home xlim but zoomed ylim, only a narrow horizontal strip of cells is "visible" — likely still ≤40
So instead of clearing, it redraws stats for that mixed view
set_ylim(original) fires → ylim_changed → nothing connected, no cleanup
The texts from step 4 are now stranded on a fully zoomed-out plot
The fix was to also connect ylim_changed to _on_zoom_change. Now after set_ylim completes, the callback fires again with both limits fully restored, finds >40 visible cells, removes all text objects, and calls draw_idle().
```

alperaltuntas (commented):
I am concerned about the complexity and performance implications of this PR compared to the value it brings.

The overlay adds subtle, largely untested logic to a UI hot path (for example the longitude wrap handling and the double-fired zoom callbacks), which is a fair amount of maintenance/testing/comprehension burden for a nice-to-have feature. Patching things with claude prompts doesn't always help with this
kind of underlying complexity, and tends to add more code around the edges
rather than reduce the overall complexity, and the result is harder to reason about and keep correct over time.

On performance, I asked Claude to profile the _on_zoom_change hot path on representative grids:

  Global quarter-degree (about 1 million cells): 18 ms per call,
  36 ms per interaction across both axes.

  Global one-tenth-degree (about 8.6 million cells): 244 ms per call,
  488 ms per interaction across both axes.

These callbacks fire repeatedly during pan and zoom drags, not just once, so interactive panning on a global grid would be visibly laggy. The check for 40 or fewer visible cells happens after all the full-grid work, so it does not protect the common zoomed-out case. Given that TopoEditor is already performance-limited, I am a bit cautious. 

So, I suggest closing this PR and revisiting it after the CrocoDash v1 release. Alternatively, you could split this PR to include only the Mean/Max/Min buttons, which I would be fine to merge before the release.

On a more general note, TopoEditor initially wasn't meant to turn into a full-blown, sophisticated GUI product. It started as a quick widget. But with these kinds of additions, we have gradually pushed it in that direction without the necessary architecture. For now, my instinct is to keep it lean and focused, and if we do want a richer interactive editor down the line, I feel like we'd need a big overhaul.

manishvenu:
This makes sense to me, what I'll do is remove the zoom functionality and keep the mean min max buttons.

alperaltuntas (approved):

PR #85, opened 2026-04-30
Set Depth from Computed Statistics

manishvenu:
Add a set depth function from computed stats


mom6_forge/topo.py
```diff
@@ -873,6 +873,30 @@ def _compute_stats(self, nx_sub, ny_sub, mask_hmin):
...
+
+        Parameters
+        ----------
+        statistic : str
+            Which depth statistic to use. Must be one of the "D_*" keys
+            in self.src.stats (e.g. "mean", "min", "max").
+        """
+
+        assert (
+            self.src.stats is not None
```
alperaltuntas: This reads `self.src.stats`, but doesn't check if `self.src` is defined. Is `self.src` always defined?

manishvenu: Yes, added.

alperaltuntas (approved):

PR #86, opened 2026-04-30
Bathymetry Pipeline Scaffolding

### LINKED ISSUE #192: Fred, Gustavo, Frank Bathy Tools/Workflow

manishvenu:
This is where I take set_from_dataset from the OLD method (Ashley Barnes regional-mom6), to the NEW workflow! 

I am happy how this turned out!

Docs: https://docs.google.com/document/d/11rGmyRcyXVK9AzdFeHzhjYPSUH9jPjWLqE35222BSkQ/edit?tab=t.0

Package Docs are going to come much later. It's very annoying to write docs when there's so many seperate PRs.

Closes https://github.com/CROCODILE-CESM/CrocoDash/issues/192


mom6_forge/topo.py
```diff
@@ -941,6 +941,88 @@ def set_depth_from_stats(self, statistic):
 
         self.send_entire_depth_change_to_tcm(self.src.stats[f"D_{statistic}"])
 
+    def diagnose_resolution(self, radius=6.371e6):
```
alperaltuntas: Please set the default radius value to _DEFAULT_RADIUS from the supergrid class.

alperaltuntas (changes requested):
Please expand the test coverage for `set_from_dataset`. Right now you are only testing a single combination (` mask_method='ocean_frac'` with `depth_method='stats'`). That leaves most of the new public surface untested. Please add tests covering:

1. The default/auto path: call set_from_dataset with mask_method=None and depth_method=None on both a fine-resolution grid and a coarse one, so the resolution-diagnostic branch selects each route automatically.
2. A test per depth_method value: 'stats', 'xesmf', and 'cressman'. 
3. A test per mask_method value: 'naturalearth', 'ocean_frac', 'dataset', and 'manual'.
4. `fill_channels=True`. At least one test that runs the workflow end-to-end with channel filling enabled, and a focused unit test of `fill_inland_lakes_and_channels` on its own.
5. Tests asserting that an unsupported mask_method and an unsupported depth_method each raise, and that the raised message lists exactly the values the code actually accepts.

alperaltuntas:
manishvenu Also, if the new tests reveal any bugs, please commit the tests first on their own. Then commit the fixes separately, with meaningful commit messages that explain what you did (i.e., no `git commit  -m "bleh" please.)

mom6_forge/topo.py

mom6_forge/topo.py
```diff
@@ -1000,9 +1085,175 @@ def set_from_dataset(
...
+                )
+            elif depth_method == "xesmf":
+                if use_stats_depth:
+                    print(
+                        "Resolution diagnostics recommend using stats-based masking, but xesmf-based depth was requested, so we will do a direct xesmf regrid anyway because it was explicitly requested, but be aware that this may lead to significant land contamination of coastal depth estimates"
+                    )
+                self.set_depth_from_xesmf(
+                    output_dir=output_dir,
+                    write_to_file=write_to_file,
+                    regridding_method=kwargs["regridding_method"],
```
alperaltuntas: This line and line 1234 will crash if `regridding_method` isn't provided in the function call. Please add `regridding_method` in the function argument list.

mom6_forge/mapping.py
```diff
@@ -10,6 +10,7 @@
 from scipy.spatial import cKDTree
 from scipy.sparse import csc_matrix, coo_matrix, csr_matrix
 import xesmf as xe
+import cf_xarray
```
alperaltuntas, line 13: Do we need to add `cf_xarray` to the environment requirements, or is it guaranteed to be included in any of the requirements?

mom6_forge/topo.py
```diff
@@ -1106,7 +1342,7 @@ def mpi_set_from_dataset(
 
             `mpirun -np NUMBER_OF_CPUS ESMF_Regrid -s bathymetry_original.nc -d bathymetry_unfinished.nc -m bilinear --src_var depth --dst_var depth --netcdf4 --src_regional --dst_regional`
 
-            4. Run Topo_object.tidy_bathymetry(args) to finish processing the bathymetry.
+            4. Run Topo_object.fill_channels() to finish processing the bathymetry.
```
alperaltuntas: update the docstring.

mom6_forge/topo.py
```diff
@@ -1037,15 +1288,8 @@ def set_from_dataset(
             If bathymetry setup fails (e.g. kernel crashes), restart the kernel and edit this cell.
             Call ``[topo_object_name].mpi_set_from_dataset()`` instead. Follow the given instructions for using mpi
             and ESMF_Regrid outside of a python environment. This breaks up the process, so be sure to call
-            ``[topo_object_name].tidy_dataset() after regridding with mpi.""")
+            ``[topo_object_name].fill_channels()`` after regridding with mpi.""")
```
alperaltuntas: update this docstring.

mom6_forge/mapping.py
```diff
@@ -943,6 +953,66 @@ def gen_rof_maps(
...
+    for cf_key in ("latitude", "longitude"):
+        try:
+            src_spatial_dims.update(input_dataset.cf[cf_key].dims)
+        except KeyError:
+            pass
+    src_spatial_dims.update(d for d in input_dataset.dims if d in _SPATIAL_NAMES)
+    if not src_spatial_dims:
+        src_spatial_dims = set(input_dataset.dims)
+    src_spatial = {
+        k: input_dataset.dims[k] for k in input_dataset.dims if k in src_spatial_dims
```
alperaltuntas: Please use `input_dataset.sizes[k]` instead of `input_dataset.dims[k] `

> xarray.Dataset.dims: ... This property will be changed to return a type more consistent with DataArray.dims in the future, i.e. a set of dimension names.
https://docs.xarray.dev/en/stable/generated/xarray.Dataset.dims.html

mom6_forge/mapping.py
```diff
@@ -943,6 +953,66 @@ def gen_rof_maps(
...
+    for cf_key in ("latitude", "longitude"):
+        try:
+            dst_spatial_dims.update(output_dataset.cf[cf_key].dims)
+        except KeyError:
+            pass
+    dst_spatial_dims.update(d for d in output_dataset.dims if d in _SPATIAL_NAMES)
+    if not dst_spatial_dims:
+        dst_spatial_dims = set(output_dataset.dims)
+    dst_spatial = {
+        k: output_dataset.dims[k] for k in output_dataset.dims if k in dst_spatial_dims
```
alperaltuntas: Please use `input_dataset.sizes[k]` instead of `input_dataset.dims[k] `

> xarray.Dataset.dims: ... This property will be changed to return a type more consistent with DataArray.dims in the future, i.e. a set of dimension names.
https://docs.xarray.dev/en/stable/generated/xarray.Dataset.dims.html

alperaltuntas (changes requested):

mom6_forge/mapping.py
manishvenu: cf_xarray is a dependency of xesmf,

mom6_forge/topo.py
manishvenu: Added

mom6_forge/topo.py

mom6_forge/topo.py

mom6_forge/mapping.py
manishvenu: Yes

mom6_forge/mapping.py
manishvenu: Done

mom6_forge/topo.py
alperaltuntas: I don't see a change here.

mom6_forge/topo.py
alperaltuntas: I don't see a change here.

manishvenu: Oops dropped this change

mom6_forge/topo.py
manishvenu: Accidentally dropped this change

mom6_forge/mapping.py
alperaltuntas: Could you add cf_xarray as a dependency in setup.py? This is in case xesmf drops it as a dependency, or if we stop depending on xesmf in the future (both of which are admittedly unlikely).

manishvenu: Done!

mom6_forge/topo.py
alperaltuntas: this print statement is still not fully fixed. you removed .mpi_set_from_dataset, but still instructing users to call it.

mom6_forge/topo.py
alperaltuntas: this print statement is still not fully fixed. you removed .mpi_set_from_dataset, but still instructing users to call it.

manishvenu: Yes, done.

mom6_forge/topo.py
manishvenu: Yes, done.

alperaltuntas (approved):

PR #92, opened 2026-05-12
ESMF to & from grid options.

manishvenu:
Changes:
Move the ESMF writing to grid with an optional mask method


mom6_forge/_supergrid.py
```diff
@@ -214,6 +237,315 @@ def to_ds(self, name=None, author: Optional[str] = None) -> xr.Dataset:
...
+        )
+
+        if mask is not None:
+            esmf_mask = mask.astype(np.int32)
+            ds["elementMask"] = xr.DataArray(
+                esmf_mask.flatten(),
+                dims=["elementCount"],
+            )
+
+        ds.to_netcdf(file_path, format="NETCDF3_64BIT")
```
alperaltuntas: Why have you dropped the disabling of _FillValue  attribute?

mom6_forge/_supergrid.py
```diff
@@ -214,6 +237,315 @@ def to_ds(self, name=None, author: Optional[str] = None) -> xr.Dataset:
 
         return ds
 
+    def to_esmf_mesh(self, file_path, mask=None, title=None, radius=_DEFAULT_RADIUS):
```
alperaltuntas: why do we have this radius argument here?

mom6_forge/_supergrid.py
```diff
@@ -214,6 +237,315 @@ def to_ds(self, name=None, author: Optional[str] = None) -> xr.Dataset:
...
+        # --- Infer grid shape from connectivity ---
+        # elementCount = ny * nx; nodeCount tells us whether cyclic or not
+        ncells = ds.dims["elementCount"]
+        nnodes = ds.dims["nodeCount"]
+        conn = ds["elementConn"].values  # (ncells, 4)
+
+        # Recover nx, ny: elementConn encodes row/col structure.
+        # For cyclic:     nnodes = nx * (ny+1)  →  nx = nnodes / (ncells/nx + 1) ... circular
+        # Simpler: scan first row of connectivity to find where the row wraps.
+        # The lower-left node index increments by 1 each element until the wrap.
```
alperaltuntas: please use already existing utility functions get_mesh_dimensions and is_mesh_cyclic_x,

alperaltuntas (changes requested):

mom6_forge/_supergrid.py
manishvenu: No reason, just removed

mom6_forge/_supergrid.py
manishvenu: Missed that, fixed.

mom6_forge/_supergrid.py
manishvenu: Done, I also replaced the grid_from_esmf_mesh with the new function in Topo all over mapping.

mom6_forge/_supergrid.py
```diff
@@ -210,6 +225,295 @@ def to_ds(self, name=None, author: Optional[str] = None) -> xr.Dataset:
...
+        """
+
+        if isinstance(file_path, xr.Dataset):
+            ds = file_path
+        else:
+            ds = xr.open_dataset(file_path)
+
+        topology = ds.attrs.get("grid_topology", None)
+        if topology == "tripolar":
+            raise NotImplementedError(
```
alperaltuntas: I am not convinced that this limitation is truly necessary. The truncation drops a set of top-row fold nodes ([: -(nx//2 - 1)]) that are duplicates of retained nodes, and topology/nx are known at read time, so it looks cleanly reversible to me.

mom6_forge/_supergrid.py
```diff
@@ -210,6 +225,295 @@ def to_ds(self, name=None, author: Optional[str] = None) -> xr.Dataset:
 
         return ds
 
+    def to_esmf_mesh(self, file_path, mask=None, title=None):
```
alperaltuntas: I think mask should be a required argument. We can perhaps have a special value "all_unmasked" where all mask values are set to 1, but even that shouldn't be a default value.

tests/test_supergrid.py
```diff
@@ -109,3 +154,266 @@ def test_uniform_spherical_supergrid():
...
+        ("non_cyclic_sg", "non_cyclic"),
+        ("cyclic_sg", "cyclic"),
+    ],
+)
+def test_roundtrip_center_coords(sg_fixture, label, request, tmp_path):
+    sg = request.getfixturevalue(sg_fixture)
+    path = tmp_path / f"{label}.nc"
+    sg.to_esmf_mesh(str(path))
+    sg2 = SupergridBase.reconstruct_from_esmf_mesh(str(path))
+    np.testing.assert_allclose(sg2.x[1::2, 1::2], sg.x[1::2, 1::2], atol=1e-10)
```
alperaltuntas: Shouldn't center coords be exactly the same?

tests/test_supergrid.py
```diff
@@ -109,3 +154,266 @@ def test_uniform_spherical_supergrid():
...
+        ("non_cyclic_sg", "non_cyclic"),
+        ("cyclic_sg", "cyclic"),
+    ],
+)
+def test_roundtrip_corner_coords(sg_fixture, label, request, tmp_path):
+    sg = request.getfixturevalue(sg_fixture)
+    path = tmp_path / f"{label}.nc"
+    sg.to_esmf_mesh(str(path))
+    sg2 = SupergridBase.reconstruct_from_esmf_mesh(str(path))
+    np.testing.assert_allclose(sg2.x[::2, ::2], sg.x[::2, ::2], atol=1e-10)
```
alperaltuntas: shouldn't corner coords be exactly the same?

tests/test_supergrid.py
```diff
@@ -109,3 +154,266 @@ def test_uniform_spherical_supergrid():
...
+    np.testing.assert_allclose(sg2.y[::2, ::2], sg.y[::2, ::2], atol=1e-10)
+
+
```
alperaltuntas, line 264: Could you add a check for edge coords too?

tests/test_topo.py
```diff
@@ -95,3 +98,47 @@ def test_topo_no_git(get_rect_topo_without_vc):
...
+    topo = get_rect_topo
+    mesh_path = str(tmp_path / "test.nc")
+    topo.write_esmf_mesh(mesh_path)
+    topo2 = Topo.from_esmf_mesh(mesh_path, git=False)
+    assert topo2.tmask.shape == topo.tmask.shape
+    assert topo2._grid.nx == topo._grid.nx
+    assert topo2._grid.ny == topo._grid.ny
+
+
+def test_topo_from_esmf_mesh_mask_roundtrip(get_rect_topo, tmp_path):
```
alperaltuntas: Why are test_topo_from_esmf_mesh_shape and test_topo_from_esmf_mesh_mask_roundtrip two separate tests? I think we should be mindful of keeping the number and runtime of added tests, (That's not to say we should reduce testing). So please try to merge tests whenever it makes sense, in this PR, and in others.

tests/test_supergrid.py
```diff
@@ -109,3 +154,266 @@ def test_uniform_spherical_supergrid():
         lon_min=0.0, len_x=10.0, lat_min=40.0, len_y=10.0, nx=nx, ny=ny
     )
     assert isinstance(sg, UniformSphericalSupergrid)
+
+
+def make_mom6_mask(ny, nx, frac_land=0.25):
```
alperaltuntas: unused

tests/test_supergrid.py
```diff
@@ -109,3 +154,266 @@ def test_uniform_spherical_supergrid():
...
+
+
+def test_cyclic_element_area_positive(cyclic_mesh):
+    assert (cyclic_mesh["elementArea"].values > 0).all()
+
+
+# ---------------------------------------------------------------------------
+# Round-trip tests
+# ---------------------------------------------------------------------------
+
```
alperaltuntas, line 244: Many of these tests share identical parametrize decorator and identical write+reconstruct setup. Could you collapse then into one smaller number of test functions wherel they are checked together? (I am not suggesting to reduce coverage, just to reduce the functions, so the tests go faster and are more tractable).

mom6_forge/mapping.py
```diff
@@ -553,8 +502,10 @@ def generate_ESMF_map_via_xesmf(
         zero out the mask in the source mesh that falls outside the rectangle defined by the destination mesh.
     """
 
-    src_grid = grid_from_esmf_mesh(src_mesh_path)
-    dst_grid = grid_from_esmf_mesh(dst_mesh_path)
+    from mom6_forge.topo import Topo
+
+    src_topo = Topo.from_esmf_mesh(src_mesh_path)
+    dst_topo = Topo.from_esmf_mesh(dst_mesh_path)
```
alperaltuntas: I vaguely remember discussing this change, but I am wondering if going from grid_from_esmf_mesh to Topo.from_esmf_mesh is really necessary. Topo.from_esmf_mesh is more heavyweight than grid_from_esmf_mesh. It reconstructs a full supergrid (interpolates u/v, recomputes dx/dy/area/rotation angles) which we don't really need, and it now fails on tripolar meshes since reconstruct_from_esmf_mesh raises error for them, so mapping a tripolar ocean grid breaks where it used to work. Also, with what you have now, the mask never reaches xe.Regridder, and doesn't get taken into account.

alperaltuntas (changes requested):
Below are some line comments.

mom6_forge/mapping.py
```diff
@@ -578,25 +529,28 @@ def generate_ESMF_map_via_xesmf(
...
+            ),
+            dims=["ny", "nx"],
         )
 
     dst_is_cyclic_x = is_mesh_cyclic_x(dst_mesh_path)
 
-    import xesmf as xe
-
+    src_ds = src_topo._grid.get_esmf_ready_tracer_ds()
+    src_ds["mask"] = src_topo.tmask
```
alperaltuntas: This adds a `mask` variable to the source grid, but the destination
grid passed to `xe.Regridder` still has none.

mom6_forge/_supergrid.py
```diff
@@ -210,6 +225,315 @@ def to_ds(self, name=None, author: Optional[str] = None) -> xr.Dataset:
...
+        # --- Recover corner (q) points from nodeCoords ---
+        node_lon = ds["nodeCoords"].values[:, 0]
+        node_lat = ds["nodeCoords"].values[:, 1]
+        axis_units = ds["nodeCoords"].attrs.get("units", "degrees")
+
+        if topology == "tripolar":
+            # Nodes stored without wrap column and with fold duplicates removed.
+            # Rows 0..ny-1 have nx nodes each; top row has nx//2+1 nodes stored
+            # (the second half were truncated as exact fold-symmetry duplicates).
+            # Recover the missing nx//2-1 nodes: qlon[ny, nx//2+j] = 360 - qlon[ny, nx//2-j].
```
alperaltuntas: This comment is inconsistent with the code.

mom6_forge/topo.py
```diff
@@ -175,6 +176,36 @@ def from_topo_file(
...
         return topo
 
+    @classmethod
+    def from_esmf_mesh(
+        cls,
+        mesh_path,
+        radius=_DEFAULT_RADIUS,
+        min_depth=0.0,
+        version_control_dir="TopoLibrary",
+        git=True,
```
alperaltuntas: Can we default `git` to `False` here?

alperaltuntas (changes requested):
There are still a couple of unresolved comments above. Below are some new ones.

manishvenu:
Hey @alperaltuntas, Sorry forgot to finish this! I'll respond to these.

mom6_forge/mapping.py
manishvenu: I moved back to the old function, but tripolar should work now.

mom6_forge/_supergrid.py
manishvenu: Done.

manishvenu:
@alperaltuntas This should be done except for the from tripolar math!

alperaltuntas (approved):
Thanks!

alperaltuntas:
@manishvenu Any reason to hold off on merging this PR?

manishvenu:
> @manishvenu Any reason to hold off on merging this PR?

I wanted to verify the tripolar stuff one last time!

PR #94, opened 2026-05-19
Revert Regional Slice Change

manishvenu:
There are a ton of issues with the regional slicer that maybe will go in the v0.1 series. I had though I fixed them in #93, but it is still wrong.


manishvenu:
I'll be merging this for now because it won't impact anything we have going on right now (Except crocontainer)

PR #95, opened 2026-05-19
Downgrade matplotlib and ipympl versions

alperaltuntas:
This is needed to make TopoEditor work on NCAR JupyterHub.

Note that Python version also needs to be downgraded from the upper limit of 3.14 to 3.13 since the working matplotlib requirement (`matplotlib>=3.9,<3.10`) isn't compatible with Python 3.14.


manishvenu (approved):
Works for me, but I wasn't able to reproduce this error!

PR #97, opened 2026-05-26
Change horizontal grid NetCDF version for efficiency when saving.

AidanJanney:
When working with high resolution grids, saving the grid can be very intensive with `NETCDF3_64BIT` format (sometimes taking 5-10 minutes). We can change this to be NETCDF4 format. This may cause issues with CESM; more testing is needed. 

Edit: Originally used compression with zlib which I thought was faster from initial testing, but standard netcdf4 format without compression is actually fastest and avoids possible errors. Saving with netcdf4 instead of netcdf3_64bit results in an order of magnitude speedup for smaller grids, and even greater savings for larger grids.


AidanJanney:
Need to test compatibility with CIME. Netcdf3 may only be needed for the ESMF mesh. We may also be able to change versions for topo and other files.

PR #98, opened 2026-05-26
Make radius optional in SupergridBase.__init__ with a warning

manishvenu:
Warn when radius is None that dx/dy/area cannot be exactly retraced. Skip writing radius to netCDF attrs if absent; load it with .get() in from_ds.

### DIFF
--- a/mom6_forge/_supergrid.py
+++ b/mom6_forge/_supergrid.py
@@ -40,7 +40,9 @@ class SupergridBase:
     def leny(self):
         return self.y.max() - self.y.min()
 
-    def __init__(self, x, y, dx, dy, area, angle_dx, axis_units, grid_type, radius=None):
+    def __init__(
+        self, x, y, dx, dy, area, angle_dx, axis_units, grid_type, radius=None
+    ):
         """
         Initialize a generic supergrid.
 
--- a/mom6_forge/_supergrid.py
+++ b/mom6_forge/_supergrid.py
@@ -40,9 +40,7 @@ class SupergridBase:
     def leny(self):
         return self.y.max() - self.y.min()
 
-    def __init__(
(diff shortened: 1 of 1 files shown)


mom6_forge/_supergrid.py
```diff
@@ -58,7 +60,15 @@ def __init__(self, x, y, dx, dy, area, angle_dx, axis_units, grid_type, radius):
...
         grid_type : str
             the type of grid being created
+        radius : float, optional
+            Sphere radius in metres. If not provided, dx/dy/area cannot be
+            exactly retraced from the stored supergrid coordinates.
         """
+        if radius is None:
+            print(
+                "No radius provided: without a stored radius, dx, dy, and area "
+                "calculations cannot be exactly retraced from the supergrid coordinates.",
```
alperaltuntas: The warning here doesn't quite hold together for me, on a couple of levels.
  
Firstly, `__init__` is a pure assignment: It stores the `dx`, `dy`, and `area` arrays the caller has already provided. Nothing in this constructor retraces metrics from coordinates, so warning about an inability to retrace at this point conflates two different responsibilities. If the concern is really about loading a legacy file with no recorded radius, the warning belongs in from_ds, not in `__init__`.

More generally though, I'd push back on the premise: the supergrid abstraction is the bundle of `x`, `y`, `dx`, `dy`, `area`, `angle_dx`. The whole point of carrying the metrics alongside the coordinates is that downstream code doesn't have to know how they were derived. In what circumstance would we actually need to retrace `dx` `dy` and `area` from x/y?

manishvenu: Yeah, that makes sense. In that sense I'll remove the radius from SuperGrid init/from_ds

alperaltuntas (approved):

PR #100, opened 2026-06-02
Move longitude_slicer to utils; fix float32 spacing tolerance

### LINKED ISSUE #99: Location of longitude_slicer
In [this commit ](https://github.com/NCAR/mom6_forge/commit/8447f1441e7dea2f0090e195c55b58cdb5523aba) [longitude_slicer](https://github.com/NCAR/mom6_forge/blob/001a64f5e71eebb421063e713f135908e95c9cce/mom6_forge/_source_bathy.py#L244) was moved out of utils and into _source_bathy. 

I don't know the context of the commit, but would there be benefit in leaving `longitude_slicer` in utils? I imagine this is a function that users may want access to outside of modifying bathymetry, so it makes sense to me to have it in a more broad `utils` file rather than a hidden one (I read the leading underscore as it being a `hidden` file not to be called directly)

Either way we need to reconcile the [import into regional-mom6 ](https://github.com/COSIMA/regional-mom6/pull/300#issuecomment-4590118531)no matter where it ends up

manishvenu:
Details:

- Moves `longitude_slicer` from `_source_bathy` to `utils`, where it belongs as a general-purpose coordinate utility
- Loosens the uniform-spacing assert to `atol=1e-4, rtol=0` to handle float32 source data (e.g. GLORYS), where accumulated rounding causes two slightly different diff values for what is conceptually a uniform grid
- Updates test import accordingly

Closes #99


ashjbarnes:
Nice I did want to revisit the tolerance too! I'll cherry pick this into my v0.0.7 PR to check it out. I found that sometimes doing computations with the len_x , xstart etc could introduce annoying rounding errors in edge cases too! Will look into it more to see if this PR fixes it :)

manishvenu:
> Nice I did want to revisit the tolerance too! I'll cherry pick this into my v0.0.7 PR to check it out. I found that sometimes doing computations with the len_x , xstart etc could introduce annoying rounding errors in edge cases too! Will look into it more to see if this PR fixes it :)


Ok great! Thanks!

ashjbarnes (approved):

alperaltuntas (approved):

PR #103, opened 2026-06-09
Add CESM integration CI for mom6_forge output files

manishvenu:
## Summary

mom6_forge generates the MOM6/CESM input files (hgrid, topog, cice_grid, esmf_mesh, scrip_grid, vgrid, chlorophyll, runoff mapping) that MOM6/CESM reads. Currently we don't check if those files run in CESM.

This PR adds CESM testing using CrocoDash functionality:

(`.github/workflows/cesm_integration.yml` + `tests/integration/`): Weekly + `workflow_dispatch` job that pulls the crocontainer (from CrocoDash PR #217), **pip-installs this specific mom6_forge commit** into the container's conda env, generates a flat-bottom test case covering all mom6_forge file types, and runs CESM for 1 model day. Requires a self-hosted runner with CESM inputdata mounted — runner label is a placeholder (`self-hosted`) to be wired up separately.


.github/workflows/cesm_integration.yml
```diff
@@ -0,0 +1,136 @@
...
+
+              # Copy case logs into scratch so they appear in artifacts
+              cp -r /workspace/case/run/ /root/cesm/scratch/case_run/ 2>/dev/null || true
+              cp -r /workspace/case/bld/ /root/cesm/scratch/case_bld/ 2>/dev/null || true
+            "
+
+      - name: Check run logs for errors
+        run: |
+          echo '--- checking run logs for errors ---'
+          RUNDIR=/home/runner/work/mom6_forge/mom6_forge/cesm_scratch/case/run
```
alperaltuntas, line 113: On line 106-107 the container copies logs to `../case_run/` and `.../case_bld/`. But RUNDIR here is `.../case/run`, and the artifact path on line 135 is also `.../case/run`.

alperaltuntas (changes requested):

PR #105, opened 2026-06-15
Add WW3 grid-input writer

alperaltuntas:
## Summary

  Adds support for generating WAVEWATCH III (WW3) grid-preprocessor input from a
  `Topo`, in support of coupling regional MOM6 with active waves. Also includes
  a few adjacent correctness fixes surfaced while wiring this up, plus a
  self-contained test suite.

  ## Changes

  ### WW3 grid-input writer (`topo.py`)
  - New `Topo.write_ww3_input(file_dir, grid_alias)` that writes the files the
    WW3 grid preprocessor expects:
    - `<alias>_x.inp` / `<alias>_y.inp` — t-point longitude / latitude
    - `<alias>_bottom.inp` — ocean depth (0 over land)
    - `<alias>_mapsta.inp` — land/sea status map from `tmask`
    - `ww3_grid.inp` — control file (CURV grid, NONE/SMPL closure from
      `is_cyclic_x`, SBF=-1 so positive depths map to negative-down elevation)
  - Depth and status maps are derived consistently from `tmask` so the two files
    never disagree.

  ### Migrate to the public `Grid.supergrid` property
  - Replace internal `self._grid._supergrid` / `self._grid.is_cyclic_x` accesses
    with the public `self._grid.supergrid[...]` API in the regridder,
    `write_scrip_grid`, and the new WW3 writer.

  ### Correctness fixes
  - `ingest_landmask`: use `self.tmask` (`self.mask` does not exist) — fixes an
    `AttributeError` in the land-mask workflow.
  - `is_rectangular`: compare lat/lon with an **absolute** degree tolerance
    (`atol`, `rtol=0`) instead of a magnitude-scaled `rtol`, since coordinates
    are angles and the acceptable deviation should not scale with magnitude.
  - Fix a copy-paste bug in the x/y-coordinate name asserts.

  ### Tests & config
  - `tests/test_ww3_input.py`: file creation, array contents/shape, control-file
    contents, and land/sea consistency.
  - `tests/test_grid.py`: curvilinear grids report non-rectangular; `atol`
    semantics.
  - `pyproject.toml`: add `[tool.pytest.ini_options]` (`testpaths`, `pythonpath`)
    so the suite runs self-contained.

  ## Limitations

  - **No wave boundary conditions yet.** There is no mechanism to impose incoming
    wave spectra at the domain edges. A regional WW3 grid generated here runs as a
    closed basin — wave energy is only what local source terms (winds) generate
    inside the domain; nothing propagates in from the boundaries.
  - **Boundary cells are treated as ordinary sea cells.** The `mapsta` status map
    is written from `tmask` as a plain land/sea mask (`1`=ocean, `0`=land). WW3's
    active-boundary code (`2`), excluded code (`3`), and ice codes (negative) are
    not emitted (there's a `TODO` for this), so edge cells are indistinguishable
    from interior ocean cells and no boundary forcing is applied.
  - **Grid representation:** emits a curvilinear (`CURV`) grid only; no
    rectilinear (`RECT`) output even for rectangular grids.
  - **Grid closure:** only open (`NONE`) and simple cyclic (`SMPL`) closures are
    derived from `is_cyclic_x`. 
  - **Hardcoded spectral grid & time steps.** Frequency/direction counts, model
    flags, and time steps in `ww3_grid.inp` are copied from the `ww3a` reference
    grid and are not yet parameterized for other resolutions or physics.
  - **Scope:** writes the WW3 *grid* preprocessor inputs only — not the MOM6-side
    wave-coupling stability settings (`WAVE_METHOD`/Stokes options), which are
    applied downstream by the case generator.

  ## Testing
  - `pytest` — full suite green (128 collected, 0 errors).


mom6_forge/topo.py
```diff
@@ -1821,6 +1821,136 @@ def write_cice_grid(self, file_path):
...
+            with open(file_dir / filename, "w") as f:
+                for j in range(ny):
+                    f.write(sep.join(fmt_cell(j, i) for i in range(nx)) + "\n")
+
+        tlon = self._grid.tlon.data  # (ny, nx), degrees
+        tlat = self._grid.tlat.data  # (ny, nx), degrees
+        # Define ocean cells from the land/sea mask so the depth and status files
+        # stay consistent even if the mask has been edited.
+        tmask = self.tmask.data  # (ny, nx), 1=ocean, 0=land
+        depth_m = np.where(tmask > 0, self._depth.data, 0.0)
```
manishvenu: Can't you use the masked_depth property for this?

tests/test_grid.py
```diff
@@ -12,10 +12,33 @@
 import pytest
 from mom6_forge.grid import Grid
 from mom6_forge.topo import Topo
+from mom6_forge._supergrid import SupergridBase
 from utils import on_cisl_machine
 import os
 
 
+def _rotated_supergrid_grid(rotation_deg, name="curv", nx=10, ny=10, d=0.1, center=10.0):
```
manishvenu: Should these functions be fixtures? Or fixture factories?

manishvenu (approved):
Don't know how the ww3 files work, but they make logical sense given the comments. I left two comments here.

tests/test_grid.py
alperaltuntas: I (and Claude) prefer to leave these as-is: These are plain helpers because they take per-call parameters (rotation angle, shear) that a plain fixture can't accept, and a fixture factory would only add a layer of indirection without any setup/teardown to justify it.

mom6_forge/topo.py
alperaltuntas: Done.

manishvenu:
Sounds good!

PR #106, opened 2026-06-23
Git efficiency for large depth/mask edits

### LINKED ISSUE #89: Better Git Efficiency w/ Full Depth Changes
I think part of the reason git is slow, especially with big domains, is the fact that the *full depth* is commited to git. That worked fine, because, ya know, it's fine, but maybe we can add a edit command that passes the full depth off to a netcdf file. Then reverse desearialize will do the same. So the valye the command will serialize is a saved file path. Then desearlaize will read from that netcdf file path. A bit tricky but we make do.

manishvenu:
## Summary

Closes #89.

- `DepthEditCommand` and `MaskEditCommand` now check the number of affected cells in `serialize()`. Below `SIZE_THRESHOLD` (10,000 cells) → existing inline JSON, unchanged. Above it → writes a content-addressed `.nc` file to `domain_dir/large_edits_optimization/` and stores only the file path. This keeps `command_history.json` tiny even for whole-domain operations on global grids (~1M cells).
- Old inline-JSON records deserialize correctly via a key check (`nc_path` vs `affected_indices`), so this is fully backward compatible.
- `.nc` files are already gitignored by the per-repo `.gitignore`, so they're never committed but persist on disk across branch checkouts.
- Adds `tests/test_git_efficiency.py` with `@pytest.mark.benchmark` tests using the tx2_3v3 global grid (960×1080) to demonstrate and guard the improvement.

## Test plan

- [x] Existing unit/integration tests pass: `pytest tests/test_topo.py tests/test_topo_command_manager.py tests/test_edit_commands.py`
- [x] Benchmark tests on GLADE: `pytest tests/test_git_efficiency.py -m benchmark -v -s`
  - Verifies `command_history.json` stays < 2 KB after `set_flat()` on tx2_3v3
  - Verifies `.nc` payload file exists at the referenced path
  - Verifies undo/redo round-trip restores depth correctly on a large domain


mom6_forge/edit_command.py
```diff
@@ -192,6 +327,20 @@ def deserialize(cls, data):
 
     @classmethod
     def reverse_deserialize(cls, data):
+        if "nc_path" in data:
+
+            def factory(topo):
+                ds = xr.open_dataset(data["nc_path"])
```
alperaltuntas: please use:
```with xr.open_dataset(...```

File: mom6_forge/edit_command.py [hunk 2]
```diff
@@ -96,20 +101,78 @@ def __call__(self):
...
+            ).to_netcdf(nc_path)
+
+        return {"type": self.__class__.__name__, "nc_path": str(nc_path)}
 
     @classmethod
     def deserialize(cls, data):
+        if "nc_path" in data:
+
+            def factory(topo):
+                ds = xr.open_dataset(data["nc_path"])
```
alperaltuntas: please use:
```with xr.open_dataset(...```

File: mom6_forge/edit_command.py [hunk 3]
```diff
@@ -119,6 +182,20 @@ def deserialize(cls, data):
 
     @classmethod
     def reverse_deserialize(cls, data):
+        if "nc_path" in data:
+
+            def factory(topo):
+                ds = xr.open_dataset(data["nc_path"])
```
alperaltuntas: please use:
```with xr.open_dataset(...```

File: mom6_forge/edit_command.py [hunk 4]
```diff
@@ -169,20 +246,78 @@ def __call__(self):
...
+            ).to_netcdf(nc_path)
+
+        return {"type": self.__class__.__name__, "nc_path": str(nc_path)}
 
     @classmethod
     def deserialize(cls, data):
+        if "nc_path" in data:
+
+            def factory(topo):
+                ds = xr.open_dataset(data["nc_path"])
```
alperaltuntas: please use:
```with xr.open_dataset(...```

File: mom6_forge/edit_command.py [hunk 5]
```diff
@@ -96,20 +101,78 @@ def __call__(self):
...
+        if not nc_path.exists():
+            xr.Dataset(
+                {
+                    "indices": xr.DataArray(idx_arr, dims=["n", "coord"]),
+                    "new_values": xr.DataArray(nv_arr, dims=["n"]),
+                    "old_values": xr.DataArray(ov_arr, dims=["n"]),
+                }
+            ).to_netcdf(nc_path)
+
+        return {"type": self.__class__.__name__, "nc_path": str(nc_path)}
```
alperaltuntas: Since `nc_path` is an absolute path, I wonder if this can lead to any portability issues. Thoughts?

mom6_forge/edit_command.py
```diff
@@ -96,20 +101,78 @@ def __call__(self):
...
+                    else None
+                ),
+            }
+
+        idx_arr = np.array(self.affected_indices, dtype=np.int32)
+        nv_arr = np.array(self.new_values, dtype=np.float64)
+        ov_arr = np.array(
+            self.old_values if self.old_values is not None else [], dtype=np.float64
+        )
+        h = hashlib.sha256(idx_arr.tobytes() + nv_arr.tobytes()).hexdigest()[:16]
```
alperaltuntas: This hash omits old_values, which means undo could restore a wrong previous state (admittedly in a very unlikely scenario). But for an abundance of caution, I suggest:

```h = hashlib.sha256(idx_arr.tobytes() + nv_arr.tobytes() + ov_arr.tobytes()).hexdigest()[:16]```

File: mom6_forge/edit_command.py [hunk 7]
```diff
@@ -169,20 +246,78 @@ def __call__(self):
...
+                    else None
+                ),
+            }
+
+        idx_arr = np.array(self.affected_indices, dtype=np.int32)
+        nv_arr = np.array(self.new_values, dtype=np.int8)
+        ov_arr = np.array(
+            self.old_values if self.old_values is not None else [], dtype=np.int8
+        )
+        h = hashlib.sha256(idx_arr.tobytes() + nv_arr.tobytes()).hexdigest()[:16]
```
alperaltuntas: This hash omits old_values, which means undo could restore a wrong previous state (admittedly in a very unlikely scenario). But for an abundance of caution, I suggest:

```h = hashlib.sha256(idx_arr.tobytes() + nv_arr.tobytes() + ov_arr.tobytes()).hexdigest()[:16]```

alperaltuntas (changes requested):

manishvenu:
@alperaltuntas All comments are finished!

alperaltuntas (approved):
LGTM. Just make sure, when resolving conflicts, testpaths/pythonpath that exist on main pyproject.toml are maintaned.

PR #107, opened 2026-06-23
Pin serial ESMF build in CI to fix intermittent SIGTERM (exit 143)

manishvenu:
Claude ran on this aand figured out the issue and it seems to work 15 times in a row!


alperaltuntas (approved):

manishvenu:
@alperaltuntas This fix for the hanging tests seems to be to make it serial so mpich is not involved

manishvenu:
Tests are not slower!

alperaltuntas (approved):

PR #109, opened 2026-06-26
Add pytest-benchmark regression gates and CI workflow

manishvenu:
## Summary

- Adds two benchmark tests to the existing test suite (`test_bench_grid_construction`, `test_bench_set_from_dataset`) using pytest-benchmark — no separate benchmarks directory needed
- Adds `benchmark.yml` CI workflow that fires on PRs to `main`, runs benchmarks, and fails if any result exceeds 110% of the committed baseline
- Normal test CI (`test.yml`) is unchanged in behaviour — `--benchmark-disable` makes benchmark fixtures a zero-overhead no-op
- Adds `tests/benchmark_baseline.json` with initial baselines generated on Derecho (AMD EPYC 9555); see the note in docs about regenerating from a GH Actions runner for cross-machine accuracy
- Adds `docs/source/benchmarks.rst` documenting the two-track design (CI gates here + SeaSloth for full characterization), how to run locally, and how to update the baseline

## Design notes

Benchmarks live in the existing test files (not a separate directory) so they share fixtures and stay in sync with the code they measure. `test_bench_set_from_dataset` uses `benchmark.pedantic` with a per-round `setup=` function to create a fresh `Topo` each round — this prevents the stats cache from making rounds 2-5 artificially fast.

## Test plan

- [ ] Verify existing tests still pass: `pytest tests/ --benchmark-disable -v`
- [ ] Verify benchmarks run and print a table: `pytest tests/ --benchmark-only -v`
- [ ] Check the `benchmark.yml` CI job passes on this PR


alperaltuntas (commented):
Looks like the latest main needs to be merged into this PR's branch and the conflicts need to be resolved.

PR #110, opened 2026-06-29
Generalize get_bounding_boxes to non-rectangular regional grids

alperaltuntas:
- Generalize get_bounding_boxes to non-rectangular regional grids. This was necessary to be able to carve out regional domains from curvilinear grids like tx2_3v2.
- Also add ww3 input write test.


manishvenu (approved):
Looks good. This was actually an issue that needed fixing for CrocoDash as well.

manishvenu:
Closes https://github.com/CROCODILE-CESM/CrocoDash/issues/234

PR #111, opened 2026-06-30
Collapse coincident pole node rows in write_esmf_mesh

alperaltuntas:
ESMF meshes written by `write_esmf_mesh` for regular lat lon grids contained coincident pole nodes (nx distinct nodes all at 90 N and S), which breaks ESMF pole-aware regridding (POLEMETHOD_ALLAVG). `write_esmf_mesh` now collapses each such pole row to a single shared node automatically, producing the topology ESMF expects.

### DIFF
--- a/mom6_forge/_supergrid.py
+++ b/mom6_forge/_supergrid.py
@@ -34,18 +34,23 @@ class SupergridBase:
 
     @property
     def is_tripolar(self):
-        nlines = 0
-        _, nx = self.x.shape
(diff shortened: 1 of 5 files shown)


alperaltuntas:
@manishvenu Could you review this PR?

manishvenu (approved):
On test, this seems to work well. I am unclear on some of the specifics but the logic makes sense and shows up in tests.

PR #112, opened 2026-07-06
Expose calendar as an argument in interpolate_and_fill_seawifs

manishvenu:
The chlorophyll climatology output hardcoded a NOLEAP calendar attribute, which is inconsistent when CrocoDash wants to match the calendar of the selected ocean forcing product.

### DIFF
--- a/mom6_forge/chl.py
+++ b/mom6_forge/chl.py
@@ -15,11 +15,10 @@ def interpolate_and_fill_seawifs(
     topo: Topo,
     processed_seawifs_path: Path | str,
     output_path: Path | str = None,
+    calendar: str = "NOLEAP",
 ):
     """
     Interpolate and fill SeaWiFS chlorophyll data to a model grid and save to NetCDF.
-
-    This function assumes a NO_LEAP calendar.
 
     This function takes gridded SeaWiFS chlorophyll data, interpolates it onto a
     super-sampled model grid, applies ocean masking, fills missing values, and writes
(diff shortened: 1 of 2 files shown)


mom6_forge/chl.py
alperaltuntas: `no_leap` and the new `calendar` arg control the same thing now, so I think `no_leap` should go. However, the values below are the only thing that sets the TIME axis, so if we delete the branch and keep the 365 day set, the file gets a calendar it doesn't match: `all_leap` would be off by a day and `360_day` by 4.5 days. Can we key the values off `calendar` instead of a bool?

alperaltuntas (changes requested):

mom6_forge/chl.py
manishvenu: I can change this, do you think we should support every netcdf calendar? or a specific set of them?

alperaltuntas: I think, for the time being, let's implement the calendar(s) we'll exercise in the workshop. And add a NotImplementedError for those we don't support yet?

manishvenu: Ok sure, and I'll move this updated code into regional_mom6.

manishvenu: For now those calendars are gregorian and no leap I believe

alperaltuntas (approved):
LGTM. Please merge main before this goes in. That should clear the CI failure.

PR #113, opened 2026-07-13
Fix polar dx antimeridian wrap

manishvenu:
Fixes dx/dy and OBC bounding-box computation for grids whose domain touches or encircles a pole

- ProjectedSupergrid.from_crs/from_center now compute dx/dy via haversine (exact great-circle distance) instead of the smallangle default — no unwrapping needed, and it stays consistent with the same sphere radius area already uses.
- get_bounding_boxes now returns the honest full longitude range (-180.0, 180.0) for an edge that crosses the antimeridian, instead of a hardcoded near-full sentinel.
- Folds in fix-dateline-wrapping-modulo's general single-seam fix (_validate_longitude_continuity, centered x re-wrapping) so both the general and pole-adjacent cases are handled by one consistent approach.
- Adds a regression test building a domain with the pole inside it, checking dx/dy directly against haversine ground truth.


mom6_forge/grid.py
```diff
@@ -425,41 +425,47 @@ def get_bounding_boxes(cls, hgrid):
...
-        }
-        north_result = {
-            "lon_min": float(hgrid.x.isel(nyp=-1).min()),
-            "lon_max": float(hgrid.x.isel(nyp=-1).max()),
-            "lat_min": float(hgrid.y.isel(nyp=-1).min()),
-            "lat_max": float(hgrid.y.isel(nyp=-1).max()),
-        }
+            assert not Grid.is_cyclic_x(hgrid)
+        else:
+            grid_check = Grid.from_supergrid_ds(hgrid)
```
alperaltuntas: This is unused. Can you remove this line alongside the assertions you commented out. (Btw, why were they commented out?)

mom6_forge/grid.py
```diff
@@ -425,41 +425,47 @@ def get_bounding_boxes(cls, hgrid):
...
+            # A raw min/max span over 180 degrees almost always means this edge
+            # crosses the antimeridian (unavoidable for a box that fully encircles
+            # a pole) rather than genuinely spanning most of the globe. Left as-is,
+            # lon_max frequently lands exactly on 180.0, which downstream lon
+            # normalization (mapping to (-180, 180]) collapses to -180.0 -- flipping
+            # the antimeridian-crossing sign check and causing the data query to
+            # grab only a thin sliver near +-180 instead of the true wraparound
+            # range. Widen explicitly to the full range instead, which downstream
+            # antimeridian-crossing handling already treats correctly.
+            if lon_max - lon_min > 180:
```
alperaltuntas: This doesn't look right at all.

mom6_forge/_supergrid.py
```diff
@@ -453,7 +459,27 @@ def from_crs(
...
+        # both of which are unavoidable for a box that fully encircles a pole,
+        # as any Arctic/Antarctic polar-cap domain does. The projected x,y grid
+        # has no such singularity: it's a plain regular Cartesian mesh, and for a
+        # conformal projection (e.g. polar stereographic) one isotropic scale
+        # factor k relates projected distance to true distance everywhere,
+        # including exactly at the pole (k is finite and smooth there).
+        dx_proj = np.diff(xx, axis=1)
+        dy_proj = np.diff(yy, axis=0)
+        k = np.asarray(Proj(crs).get_factors(lon, lat).parallel_scale)
+        sg.dx = dx_proj / (0.5 * (k[:, :-1] + k[:, 1:]))
```
alperaltuntas: This is wrong. k comes from the CRS, so dx and dy are now ellipsoidal (WGS84) while area is still computed on the sphere of radius radius.

alperaltuntas (changes requested):

mom6_forge/grid.py
manishvenu: Yes removed.

mom6_forge/_supergrid.py
manishvenu: So instead of doing it this way, I changed to using the haversine method, which would work just as well.

manishvenu:
@alperaltuntas I realized #120 was very similar to this PR so i merged the fix there, where we use the MOM6 function `module_around_point`instead of the half-wraparound here , and using haversine instead of the complicated scaling in projected setup. I think this simplifies this PR pretty drastically which is great, but is definetely a change. What it adds is a check to make sure we are not "jumping in longitude", so no 179, 180, -179 at the seams for example, whcih solves a lot of our problems like this.

mom6_forge/grid.py
manishvenu: Replaced this with a module_around_point rotator, so this should be better now.

mom6_forge/_supergrid.py
```diff
@@ -173,6 +219,14 @@ def _init_from_xy(
         # Clamp to valid geographic range (floating-point overshoot from projection in some cases)
         y = np.clip(y, -90.0, 90.0)
 
+        # Fix raw dateline-wrap discontinuities in x before differencing for dx/dy/area/angle.
+        # Skipped near poles, where every longitude legitimately converges.
+        if np.abs(y).max() < SupergridBase._POLE_ADJACENT_LAT:
+            center_lon = x[x.shape[0] // 2, x.shape[1] // 2]
+            if _max_adjacent_diff(x) > 180.0:
+                x = modulo_around_point(x, center_lon, 360)
+            x = x - np.floor(center_lon / 360) * 360
```
alperaltuntas: This line sits outside the `_max_adjacent_diff(x) > 180` guard above, so it runs on
every non-polar grid, including ones that never had a seam to begin with.

mom6_forge/_supergrid.py
```diff
@@ -918,9 +972,9 @@ def expand(self, n_cells=1) -> "SupergridBase":
         for _ in range(2 * n_cells):
             padded = self._create_expanded_supergrid(x, y)
             x, y = padded.x.values, padded.y.values
-        assert -90 <= y.min() and y.max() <= 90, (
-            "Expanded supergrid exceeds ±90 degrees latitude; check the input grid and expansion width."
-        )
+        assert (
+            -90 <= y.min() and y.max() <= 90
+        ), "Expanded supergrid exceeds ±90 degrees latitude; check the input grid and expansion width."
         return type(self)._init_from_xy(x, y, grid_type=self.grid_type)
```
alperaltuntas: This drops both the radius and `dx_dy_calc_type`, so the expanded grid is rebuilt with
the defaults rather than with whatever the parent was built with. For a polar
ProjectedSupergrid that puts back exactly the bug this PR fixes

mom6_forge/_supergrid.py
```diff
@@ -88,6 +100,40 @@ def __init__(self, x, y, dx, dy, area, angle_dx, axis_units, grid_type):
...
+        # can't fall outside 180 degrees either side of that: [-180, 540).
+        low_bound, high_bound = -180.0, 540.0
+        if x.max() > high_bound + tol or x.min() < low_bound - tol:
+            raise ValueError(
+                f"Longitude values (range [{x.min():.4f}, {x.max():.4f}]) fall "
+                f"outside [{low_bound}, {high_bound}]; x looks unwrapped/"
+                "unnormalized rather than a valid geographic longitude."
+            )
+        max_jump = _max_adjacent_diff(x)
+        if max_jump > max_adjacent_jump:
```
alperaltuntas: This makes existing supergrid files unreadable.

mom6_forge/grid.py
```diff
@@ -412,36 +413,34 @@ def get_bounding_boxes(cls, hgrid):
...
+            hgrid.x.isel(nxp=-1).values, hgrid.y.isel(nxp=-1).values
+        )
+        west_result = _lon_lat_bounds(
+            hgrid.x.isel(nxp=0).values, hgrid.y.isel(nxp=0).values
+        )
+        south_result = _lon_lat_bounds(
+            hgrid.x.isel(nyp=0).values, hgrid.y.isel(nyp=0).values
+        )
+        north_result = _lon_lat_bounds(
+            hgrid.x.isel(nyp=-1).values, hgrid.y.isel(nyp=-1).values
```
alperaltuntas, line 448: This fallback still trips on a genuinely wide domain though: a non-cyclic grid spanning -170 to 30 at the equator has a real 200 degree span, so `ic` and `south` still come back as the whole globe.

alperaltuntas (changes requested):
More line comments below though. The new issues are all in the expand-supergrid
and dateline commits rather than in the polar fix itself, so I'd suggest we
either drop this PR from v1.0, or cut it back to just the polar dx fix and let
#122 and the dateline work land separately. Between the number of commits, the
size of the changes, and the bugs that keep turning up, I am not confident about
this PR ahead of the workshop.

mom6_forge/grid.py
manishvenu: I think this is better check here, where only if the lon bounds include a pole do we give up and send back the globe.

mom6_forge/_supergrid.py
manishvenu: I removed the validation check for now.

mom6_forge/_supergrid.py
manishvenu: Fixed this by adding both of these as class attributes. I didn't fully close the roundtrip of writing a ds reading from a ds so that if we get a grid that doesn't include the attributes it's ok.

mom6_forge/_supergrid.py
manishvenu: Moved into.

manishvenu (commented):
I decided to remove the validation, and now it's basically three changes. get_bounding_boxes, supergrid class args, and I added a small bug that showed up in longitude_slicer if you request more than the entire globe.

manishvenu:
Haven't run through the tutorial example, but will this afternoon.

alperaltuntas (approved):

PR #114, opened 2026-07-14
Add boundaries to WW3 Mapsta file.

manishvenu:
Changes:

1. New function to do the mapsta, where if it's not cyclic the east/west boundaries become boundaries. North/South are always boundaries. 
2. Add a little commenting because the mapsta file should be in the input boundaries section of ww3_grid.inp not part of the other 3 files.

### DIFF
--- a/mom6_forge/topo.py
+++ b/mom6_forge/topo.py
@@ -1997,6 +1997,50 @@ class Topo:
             format="NETCDF3_64BIT",
         )
 
+    def _compute_ww3_mapsta(self):
+        """
+        Compute the WW3 mapsta status-code array for this grid.
+
+        Status codes follow WW3's convention: 0 = land, 1 = interior sea
+        point, 2 = active boundary point. Ocean cells on the grid perimeter
+        are flagged as active boundary points; ww3_grid promotes exactly
+        these points to boundary points via the "Input boundary points"
+        list in ww3_grid.inp (see write_ww3_input). A reentrant (cyclic-x)
+        edge has no physical boundary, so the east/west edges are only
+        flagged when the grid is not cyclic in x. Tripolar grids are
+        rejected outright (see assertion below): write_ww3_input's closure
+        logic has no 'TRPL' case, so the north/south edges are always
+        treated as physical boundaries here, which would be wrong for a
+        tripolar grid's pole fold.
+
+        Raises
+        ------
+        AssertionError
+            If the grid is tripolar.
+
+        Returns
+        -------
+        numpy.ndarray
+            (ny, nx) integer array of WW3 mapsta status codes.
+        """
+        assert not self._grid.is_tripolar(self._grid._supergrid), (
+            "write_ww3_input does not support tripolar grids: the pole fold "
+            "would be misidentified as an open boundary."
+        )
+
+        tmask = self.tmask.data  # (ny, nx), 1=ocean, 0=land
+        mapsta = tmask.astype(int).copy()
+
+        is_boundary = np.zeros_like(mapsta, dtype=bool)
+        is_boundary[0, :] = True
+        is_boundary[-1, :] = True
+        if not self._grid.supergrid.is_cyclic_x:
+            is_boundary[:, 0] = True
+            is_boundary[:, -1] = True
+
+        mapsta[is_boundary & (tmask == 1)] = 2
+        return mapsta
+
     def write_ww3_input(self, file_dir, grid_alias):
         """
         Write the text-based WW3 input files ww3_grid.inp, [grid_alias]_x.inp, [grid_alias]_y.inp,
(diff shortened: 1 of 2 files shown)


mom6_forge/topo.py
```diff
@@ -1997,6 +1997,50 @@ def write_cice_grid(self, file_path):
...
+        assert not self._grid.is_tripolar(self._grid._supergrid), (
+            "write_ww3_input does not support tripolar grids: the pole fold "
+            "would be misidentified as an open boundary."
+        )
+
+        tmask = self.tmask.data  # (ny, nx), 1=ocean, 0=land
+        mapsta = tmask.astype(int).copy()
+
+        is_boundary = np.zeros_like(mapsta, dtype=bool)
+        is_boundary[0, :] = True
```
alperaltuntas: This unconditionally sets boundaries to having prescribed b.c., which isn't something we always want. The east/west edges are already skipped for cyclic-x, but the north/south edges are still flagged even for a global grid, where prescribed b.c. at the poles doesn't mean much. Even in regional domains, the users may not always mark a particular edge for b.c. So, I think, rather than unconditionally setting all boundaries to receiving b.c., there should be flags to determine which edges should be configured for b.c. (Then, in CrocoDash/visualCaseGen, these flags can be set accordingly.) Also, we should have a sanity check to confirm flags aren't set for east/west edges.

mom6_forge/topo.py
```diff
@@ -1997,6 +1997,50 @@ def write_cice_grid(self, file_path):
...
+        ------
+        AssertionError
+            If the grid is tripolar.
+
+        Returns
+        -------
+        numpy.ndarray
+            (ny, nx) integer array of WW3 mapsta status codes.
+        """
+        assert not self._grid.is_tripolar(self._grid._supergrid), (
```
alperaltuntas: Can we move the check up next to the `assert "degrees" in ...` at the top of `write_ww3_input`? Also, this uses `self._grid._supergrid` while line 2037 just below
uses the `supergrid` property.

mom6_forge/topo.py
```diff
@@ -1997,6 +1997,50 @@ def write_cice_grid(self, file_path):
             format="NETCDF3_64BIT",
         )
 
+    def _compute_ww3_mapsta(self):
+        """
+        Compute the WW3 mapsta status-code array for this grid.
+
+        Status codes follow WW3's convention: 0 = land, 1 = interior sea
+        point, 2 = active boundary point. Ocean cells on the grid perimeter
+        are flagged as active boundary points; ww3_grid promotes exactly
```
alperaltuntas: There is no "Input boundary points" list in the file we write. The mapsta file is declared with FROM='NAME' at line 2168, so ww3_grid reads the codes verbatim, and the segment list is only read when FROM='PART'. Can we reword this to say ww3_grid keeps
the 2s and demotes any perimeter 1 to 3?

mom6_forge/topo.py
```diff
@@ -2118,6 +2158,13 @@ def _write_rows(filename, fmt_cell, sep):
             f.write(
                 f"  -0.1 {self._min_depth:.2f} 23 -1. 1 1 '(....)' 'NAME' '{bottom_file}'\n"
             )
+            f.write(
+                "$ Input boundary points and excluded points -------------------------- $\n"
+                "$ The first line identifies where to get the map data, by unit number\n"
+                "$ IDLA and IDFM, format for formatted read, FROM and filename\n"
+                "$ if FROM = ’PART’, then segmented data is read from below, else\n"
```
alperaltuntas: Why do we use curly quotes here, instead of plain `'`?

mom6_forge/topo.py
```diff
@@ -1997,6 +1997,50 @@ def write_cice_grid(self, file_path):
...
+        numpy.ndarray
+            (ny, nx) integer array of WW3 mapsta status codes.
+        """
+        assert not self._grid.is_tripolar(self._grid._supergrid), (
+            "write_ww3_input does not support tripolar grids: the pole fold "
+            "would be misidentified as an open boundary."
+        )
+
+        tmask = self.tmask.data  # (ny, nx), 1=ocean, 0=land
+        mapsta = tmask.astype(int).copy()
```
alperaltuntas: `.copy()` is redundant.

alperaltuntas (changes requested):

mom6_forge/topo.py
manishvenu: Copied from another web source I''ll remove this

mom6_forge/topo.py
manishvenu: Yup

mom6_forge/topo.py
manishvenu: I think the trick here is that the mapsta is written at grid generation time and not forcing time, so we have no knowledge on user requested boundaries.

manishvenu: When this gets merged, I'll move the ww3 grid generation to the ww3 configure so that we can boundary options.

alperaltuntas (approved):

PR #116, opened 2026-07-16
patch overflow in xarray when saving netcdf

ashjbarnes:
the old fill value was causing an overflow in xarray

Note that this change will probably make rmom6 incompatible with FRE tools moving forwards but I'll need to test that


manishvenu (approved):
Change Version!

PR #118, opened 2026-07-20
Allow separate resolution_lat/resolution_lon for uniform_spherical Grid

manishvenu:
Adds resolution_lat and resolution_lon as an alternative to the single resolution argument, so nx/ny can be derived from independent per-axis degree spacing instead of forcing lenx and leny to share one resolution.


ashjbarnes (approved):
works well!

PR #119, opened 2026-07-20
Ensure grids don't have discontinuities in longitude AND max lon < 360

ashjbarnes:
Grid generation in mom6_forge can currently do two problematic things:

1. Generate grids where x jumps suddenly from 180 to -180
2. Longitudes can be larger than 360 degrees

The former problem only occurs with _projected_ grids, rather than grids that are populated with a linspace or arange command, like `rectilinear_cartesian` and `uniform_spherical`. The proposed fix for the projected grids is to move the prime meridian of the destination grid to the domain's centre, then perform the projection. Afterwards, shift the longitude values again so that zero longitude lines up with Greenwich 

This can introduce cases where longitude > 360 in the projected grids as well as the simpler types. In all cases, the suggested fix is to subtract 360 if the largest longitude is greater than 360. In effect this means that [350:370] -> [-10,20]. 

This grid behaviour may work with mom6 ( @manishvenu is testing it) but I think having a discontinuity in a coordinate would create headaches for users analysing and plotting output down the track.

### DIFF
--- a/mom6_forge/_supergrid.py
+++ b/mom6_forge/_supergrid.py
@@ -341,6 +341,11 @@ class UniformSphericalSupergrid(SupergridBase):
         grid_y = lat_min + jindp * len_y / ny_total  # latitude edges
         grid_x = lon_min + iindp * len_x / nx_total  # longitude edges
 
+        # ---------------------------------------------------------------------
+        # If there are longitudes larger than 360, shift to -180 -> 180 instead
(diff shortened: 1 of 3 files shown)


ashjbarnes:
Things to check: I've tried to find all the instances of longitude specification in the _supergrid class. I may have missed some if longitude is specified elsewhere?

manishvenu:
Clsoed in favor of #120

PR #120, opened 2026-07-21
dateline wrapping by modulo

manishvenu:
There is an issue with seams me and @ashjbarnes were trying to fix: 

We can't have a major jump in longitude across seams, it messes with the metric calculations. So we use degrees east at the antimeridian, and negative degrees (-10 ->10) at the prime meridian. The way that's done is with the module_around_point function.


ashjbarnes (approved):
works on my test domains

ashjbarnes:
@manishvenu just checking you're not waiting on me to merge this?

manishvenu:
No!

manishvenu:
Superseded — folded into #113 (fix-polar-dx-antimeridian-wrap), which now merges this branch's general single-seam fix together with the pole-encircling fix, all in one place.

PR #121, opened 2026-07-21
replace global hardcoded plotting function with regional one

ashjbarnes:
I've replaced the old plotting functions which @manishvenu said are unused with a regional one. If we still need the global ones for whatever reason we could add an easy check at the start of .plot() that calls _plot_regional() if the domain isn't global


ashjbarnes:
ok idk why this broke the CI - there aren't even tests for the old plot functions I deleted

manishvenu:
There's a bug in the testing suite! Nothing to do with this PR!

PR #122, opened 2026-08-06
Add SupergridBase.expand() for growing a supergrid by a halo

manishvenu:
Needed by CrocoDash's upcoming CICE restoring-file generation, which requires the case's domain plus a multi-cell halo on every side.


mom6_forge/_supergrid.py
```diff
@@ -902,6 +902,19 @@ def _create_expanded_supergrid(x, y, expansion_width=1) -> xr.Dataset:
         )
         return pseudo_supergrid
 
+    def expand(self, n_cells=1) -> "SupergridBase":
```
alperaltuntas, line 905: Please add a validation (>=1) for ncells

mom6_forge/_supergrid.py
```diff
@@ -902,6 +902,19 @@ def _create_expanded_supergrid(x, y, expansion_width=1) -> xr.Dataset:
...
+        """
+        Return a new supergrid padded by ``n_cells`` T-cells of linear
+        extrapolation on every side (a halo), by repeatedly applying
+        :func:`_create_expanded_supergrid` (each call pads by half a T-cell
+        per side) and rebuilding full grid metrics from the resulting x/y.
+        """
+        x, y = self.x, self.y
```
alperaltuntas, line 917: Cyclic x and tripolar grids lose their topology: for a 0-360 grid `is_cyclic_x` flips to
False and `lenx` becomes 400, which then trips the `assert 0 < lenx <= 360.0` in
`grid.py`. Can we raise `NotImplementedError` for `is_cyclic_x` and `is_tripolar` for now?

mom6_forge/_supergrid.py
```diff
@@ -902,6 +902,19 @@ def _create_expanded_supergrid(x, y, expansion_width=1) -> xr.Dataset:
...
+        for _ in range(2 * n_cells):
```
alperaltuntas, line 918: For a domain that already reaches 90N, this walks the halo rows past the pole, and
`_init_from_xy` then clips them back to 90. We should detect it and probably raise an error (unless you can come up with a valid solution).

mom6_forge/_supergrid.py
```diff
@@ -902,6 +902,19 @@ def _create_expanded_supergrid(x, y, expansion_width=1) -> xr.Dataset:
...
+            padded = self._create_expanded_supergrid(x, y)
```
alperaltuntas, line 919: `_create_expanded_supergrid` already takes an `expansion_width` arg that it rejects for
anything but 1. Would it be cleaner to implement that as one vectorized extrapolation,
rather than calling the helper 2*n_cells times? That would also avoid the repeated
full-array copies, which I'd worry about at 1/12 degree. The interior `angle_dx` is
unchanged by the padding, so recomputing angles over the whole grid is avoidable too.

mom6_forge/_supergrid.py
```diff
@@ -902,6 +902,19 @@ def _create_expanded_supergrid(x, y, expansion_width=1) -> xr.Dataset:
...
+            x, y = padded.x.values, padded.y.values
+        return SupergridBase._init_from_xy(x, y, grid_type=self.grid_type)
```
alperaltuntas: Can we use `type(self)._init_from_xy(...)`? Right now `UniformSphericalSupergrid.expand(2)` returns a bare `SupergridBase` while
`grid_type` still says "uniform_spherical".

tests/test_rotation.py
```diff
@@ -198,6 +198,22 @@ def test_expanded_supergrid_generation(get_curvilinear_supergrid):
     return
 
 
+def test_expand_adds_halo_and_preserves_interior(get_curvilinear_supergrid):
```
alperaltuntas: This only checks shape and that the interior x/y are preserved, so none of the issues I
flagged in `_supergrid.py` would fail it. Can we add a near-pole case and assert on
dx/dy/area and `grid_type` as well?

alperaltuntas (changes requested):

mom6_forge/_supergrid.py
manishvenu: Yes

mom6_forge/_supergrid.py
manishvenu: Yup

mom6_forge/_supergrid.py
manishvenu: For now, I'll raise an error.

mom6_forge/_supergrid.py
manishvenu: Yes

tests/test_rotation.py
manishvenu: yes

mom6_forge/_supergrid.py
manishvenu: I was thinking about this. It's a somewhat big change?

manishvenu: So maybe, I'll hold off on this for now, and make an issue?

alperaltuntas (approved):
After the black failure is fixed, this PR can be merged.

PR #126, opened 2026-08-19
Fix tarea double-counting a supergrid quadrant

manishvenu:
## Problem

`Grid.tarea` ([grid.py:834-839](https://github.com/NCAR/mom6_forge/blob/main/mom6_forge/grid.py#L834-L839)) sums:

```python
sg.area[::2, ::2] + sg.area[1::2, 1::2] + sg.area[::2, 1::2] + sg.area[::2, 1::2]
```

`sg.area[::2, 1::2]` appears twice and `sg.area[1::2, ::2]` never appears, so each T-cell's area is three of its four supergrid quadrants with one of them counted twice.

This is wrong on **every** grid, not just exotic ones. On the existing `panama1` fixture (0.1°, 278–282°E, 7–10°N):

- max per-cell relative error **1.58e-04**
- domain total off by **0.014%**

The error scales with cell size and with latitude, because what the duplication misrepresents is the north–south variation in cell area within a T-cell. It is not a rounding artefact — the corrected sum reproduces the block sum to 1.2e-16.

`tarea` is exported as the conservative-regridding weight in `to_esmf_ds()`, so the error propagates into forcing generation rather than staying a diagnostic curiosity.

## Fix

Sum the four distinct quadrants.

## Tests

Two regression tests in `tests/test_grid.py`:

- `test_grid_tarea_sums_all_four_supergrid_quadrants` — compares against the `(ny, 2, nx, 2)` block sum of `sg.area`.
- `test_grid_tarea_matches_analytic_spherical_area` — compares the domain total against the closed-form spherical-box area, which does not go through `sg.area` at all, so a future bug in the supergrid areas cannot hide behind a matching bug in the quadrant bookkeeping.

`tests/test_grid.py tests/test_supergrid.py tests/test_mapping.py tests/test_masks.py tests/test_rotation.py tests/test_utils.py` → 90 passed. The one failure, `test_is_tripolar`, fails identically on `main` (unrelated, `AttributeError: 'Dataset' object ...`).

## Relationship to #113

Independent — #113 touches `_calc_dx_dy` and `get_bounding_boxes`, this touches only the `tarea` sum. Both were surfaced by the same domain-sweep work in CrocoDash; no conflicts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mom6_forge/grid.py
+++ b/mom6_forge/grid.py
@@ -830,12 +830,14 @@ class Grid:
             dims=["ny", "nx"],
             attrs={"name": "angle q-grid makes with latitude line", "units": "degrees"},
         )
-        # T area
+        # T area: each T-cell spans a 2x2 block of supergrid cells, so its area
+        # is the sum of all four quadrants -- lower-left, lower-right, upper-left
+        # and upper-right.
         self.tarea = xr.DataArray(
             sg.area[::2, ::2]
-            + sg.area[1::2, 1::2]
             + sg.area[::2, 1::2]
-            + sg.area[::2, 1::2],
+            + sg.area[1::2, ::2]
+            + sg.area[1::2, 1::2],
             dims=["ny", "nx"],
             attrs={"name": "area of t-cells", "units": "meters^2"},
         )
--- a/tests/test_grid.py
+++ b/tests/test_grid.py
@@ -12,7 +12,7 @@ import xarray as xr
 import pytest
 from mom6_forge.grid import Grid
 from mom6_forge.topo import Topo
-from mom6_forge._supergrid import SupergridBase
+from mom6_forge._supergrid import SupergridBase, _DEFAULT_RADIUS
 from utils import on_cisl_machine
 import os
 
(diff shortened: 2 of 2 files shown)


manishvenu:
Validated against #113 as well, since the two touch adjacent code.

`ci-integration-metrics` is a throwaway branch merging this PR into #113. The merge needed one trivial resolution — both sides had added to the same two import lines in `_supergrid.py`, so I took the union. mom6_forge's own suite on that branch: **90 passed**, with only the pre-existing `test_is_tripolar` failure (which fails identically on `main`).

I also ran CrocoDash's 17-domain grid/bbox sweep ([CrocoDash#274](https://github.com/CROCODILE-CESM/CrocoDash/pull/274)) against it, which is what surfaced both of these bugs:

| `mom6_forge` | result |
|---|---|
| current CrocoDash pin | 209 passed, 6 skipped, 27 xfailed |
| `ci-integration-metrics` (#113 + this) | 229 passed, 6 skipped, 7 xfailed |

No failures either way — the 20 xfails that flip are exactly the `tarea` and `dx`/`dy` ones these two PRs fix. CrocoDash decides those with a runtime capability probe rather than a version pin, so nothing downstream breaks when either lands.

One note on the remaining 7: they're the globally-cyclic grid (which `get_bounding_boxes` rejects outright) plus the inflated bounding box on a rotated domain crossing the antimeridian. **#113 doesn't fix that last one** — it widens any >180° raw span to the full range rather than narrowing it, so a ~400 km domain still reports 359.999°. That looks deliberate given the `lon_max == 180.0` → `-180.0` collapse it avoids, and narrowing properly would need the box to carry a wrapped range (`lon_min > lon_max`) that every consumer would have to understand. Flagging it rather than proposing a change here, since it's out of scope for this PR.

PR #127, opened 2026-08-25
Get mesh grid dims directly in write_esmf_mapping_file

manishvenu:
`write_esmf_mapping_file` only needs each mesh's `(ny, nx)` shape, but was calling `Topo.from_esmf_mesh()` twice to get it — reconstructing full supergrid geometry as a side effect.

Now uses `get_mesh_dimensions()` directly, and `src_grid_dims`/`dst_grid_dims` plus the `ni_*`/`nj_*` index vectors are built from those ints.

`tests/test_mapping.py`: 7 passed.


alperaltuntas (approved):
Awesome, thanks!

PR #137, opened 2026-09-03
Coastline-masked, topography-aware runoff smoothing, and performance improvements

alperaltuntas:
Smooths runoff mapping weights along coastal topology instead of radially, and make generating the maps efficient at high resolution.

## Changes
 - generate_ESMF_map_via_xesmf gains coastline_masking, so nearest-neighbor mapping targets only ocean cells adjacent to the coast.
 - compute_smoothing_weights walks 8-connectivity through ocean cells (BFS), accumulating distance hop by hop, instead of a KD-tree ball query. Land is a barrier, so runoff no longer smears across peninsulas and isthmuses. 
 - gen_rof_maps verifies both maps conserve runoff before finishing.

## New performance:

JRA rof mesh mapped to various ocn grids:

| ocn grid | main | + algorithm | + perf work | total |
|----------|-------:|-----------:|-----------:|------:|
| gx1v7 (500 km)  |  14.68s |   9.86s |  1.80s |  8.2x |
| tx0.25 (100 km) |  86.69s |  23.71s |  5.48s | 15.8x |
| tx0.1 (20 km)   | 214.41s |  81.42s | 12.85s | 16.7x |
| **total**       | **315.78s** | **114.99s** | **20.13s** | **15.7x** |

The algorithmic changes account for 2.7x: coastline masking restricts smoothing to coastal cells, and the BFS replaces a query_ball_tree over the whole mesh. The remaining 5.7x is from vectorizing the neighbor adjacency and vertex-coordinate construction, caching cell areas per mesh file rather than recomputing them five times per call, and writing  NETCDF4_CLASSIC instead of netCDF3.

## File format: 
Mapping files are now written as NETCDF4_CLASSIC rather than 64-bit offset. netCDF3 cannot store a non-final variable of 4 GiB or more, which a JRA -> tx0.25 map exceeds at 1.76 billion weights. NETCDF4_CLASSIC has no such limit and is faster to write at every size. CESM reads these files only through ESMF_FieldSMMStore (cmeps/mediator/med_map_mod.F90); ESMF 8.9.1, the version Derecho loads, reads both formats, verified on Casper. The netCDF-4 refusal in the old SVN rimport went away with the new inputdata tooling, which has published netCDF-4 runoff mapping files since October 2025.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,18 @@ jobs:
         run: |
           git config --global user.email "runner@example.com"
           git config --global user.name "CI Runner"
+
+      # CESM input data (ESMF meshes) used by the mapping integration tests.
+      # ~28 MB, fetched once and reused; the key changes only when the file
+      # manifest in tests/utils.py does.
+      - name: Cache CESM input data
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/mom6_forge/inputdata
+          key: cesm-inputdata-${{ hashFiles('tests/utils.py') }}
+          restore-keys: |
+            cesm-inputdata-
+
       - name: Run pytest
         shell: bash -l {0}
         run: |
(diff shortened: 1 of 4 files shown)


manishvenu (approved):
I'm not going to pretend to truly get this PR but BFS makes sense with neighbors being discovered. Very neat that solves the peninsula problem/what Frank was seeing last year.

PR #138, opened 2026-09-13
Fix plotting in TopoEditor

manishvenu:
## Summary
- Grids/domains straddling the antimeridian were plotting (or zooming to) the entire globe instead of the actual domain in both `GridCreator` and `TopoEditor`.

This is based on `fix-polar-dx-antimeridian-wrap` rather than `main`, since that branch already guarantees the underlying supergrid longitude data has no raw seam jumps — this PR only needed to fix the display/interaction layer (`grid_creator.py`, `topo_editor.py`), not the grid data itself.


manishvenu:
@alperaltuntas Running the tutorial domain before this PR with the topo editor spans the entire globe instead of the domain we want.

